PR #12, opened 2020-08-25
Turn on strict null checks and associated fixes

eyingxuan:
[ch2288] and typescript null checks are turned on!


eyingxuan:
Remember to not redirect to 404 is it's an internet error

PR #14, opened 2020-08-27
Order_by queue view

eyingxuan:
[ch2317]

### DIFF
--- a/backend/ohq/views.py
+++ b/backend/ohq/views.py
@@ -271,11 +271,15 @@ class QueueViewSet(viewsets.ModelViewSet):
             .values("count")
         )
 
-        return Queue.objects.filter(course=self.kwargs["course_pk"], archived=False).annotate(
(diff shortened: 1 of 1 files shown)


backend/ohq/views.py
```diff
@@ -271,10 +271,14 @@ def get_queryset(self):
...
-            questions_asked=Subquery(questions_asked[:1]),
-            staff_active=Subquery(staff_active[:1]),
+        return (
+            Queue.objects.filter(course=self.kwargs["course_pk"], archived=False)
+            .annotate(
+                questions_active=Subquery(questions_active[:1], output_field=IntegerField()),
+                questions_asked=Subquery(questions_asked[:1]),
+                staff_active=Subquery(staff_active[:1]),
+            )
+            .order_by()
```
davish, line 281: This will implicitly order by primary key, i'm pretty sure. We probably would like to explicitly order by `time_asked` or whatever that field is called. But it should work!

PR #23, opened 2020-08-31
Last question card

ArmaanT:
This PR moves the last question card to be specific to a queue, and fixes its functionality.

### DIFF
--- a/frontend/components/Course/StudentQueuePage/StudentQueue.tsx
+++ b/frontend/components/Course/StudentQueuePage/StudentQueue.tsx
@@ -146,13 +146,13 @@ const StudentQueue = (props: StudentQueueProps) => {
                         toastFunc={updateToast}
                     />
                 )}
-                <Grid.Row columns={1}>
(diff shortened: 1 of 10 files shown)


frontend/components/Course/StudentQueuePage/StudentQueue.tsx
```diff
@@ -136,18 +146,13 @@ const StudentQueue = (props: StudentQueueProps) => {
...
-                    questions.length === 0 && (
-                        <Message
-                            style={{ marginTop: "10px" }}
-                            info
-                            header="Question already in queue"
-                            icon="comment alternate outline"
-                            content="You already have asked a question in another queue"
-                        />
-                    )} */}
+                <Grid.Row columns={1}>
```
eyingxuan: Can you wrap the entire row in a lastQuestions check here? This will prevent an empty row being rendered with nothing

ArmaanT: resolved

PR #33, opened 2020-09-12
Websocket Integration

eyingxuan:


frontend/hooks/data-fetching/course.ts
```diff
@@ -1,4 +1,11 @@
+import { useEffect } from "react";
 import useSWR, { mutate as globalMutate } from "swr";
+import {
+    useRealtimeResourceList,
+    useRealtimeResource,
+} from "@pennlabs/rest-live-hooks";
+// TODO: REMOVE THIS AS SOON AS WE REFACTOR
```
eyingxuan, line 7: Just wanted to point this out, I think this is the best solution for now - I don't want to spend time refactoring out everything and delaying websockets, but their types don't coerce well so I have to import this version

frontend/components/Course/StudentQueuePage/QuestionForm.tsx
```diff
@@ -1,14 +1,10 @@
 import React, { useState } from "react";
 import { Segment, Form, Header, Button } from "semantic-ui-react";
+import { mutateResourceListFunction } from "@pennlabs/rest-hooks/dist/types";
```
eyingxuan, line 3: @davish Should we export types from the library? Or did you choose not to export this for good reason?

davish: yeah, we should. I wish there was a way for us to alias this to `rest-hooks/types` but I'll put it in the index exports in the next release (wow that sounds so professional)

eyingxuan:
Things to polish
- [ ] Avoid subscriptions when queue is closed (Comment: I'm not sure how to do this as of now - you can't wrap a hook in a conditional, and I'm not sure what effect passing undefined to the realtime hook will affect the logic)
- [x] Add orderby for queue resource list
- [ ] Figure out proxy https://github.com/chimurai/http-proxy-middleware/blob/master/recipes/router.md
- [x] Prefix all routes with API instead of using global provider so the POST/PATCH/PUTS will play nice

eyingxuan:
Leaving some notes here for upstream ~frontend-core bugs~ I agree that this should not be the default behavior
- [x] An UPDATED action might have to trigger a revalidate (updating a question status might cause it to disappear from the view): this should probably be an option

ArmaanT (approved):
This approval was made under duress

PR #35, opened 2020-09-18
Course title search + question export

ArmaanT:
This PR:
* Displays and allows students to search for courses by their title
* Configures the backend to export the question summary view to an excel file. The frontend **has not** been configured to support this yet
image

### DIFF
--- a/backend/ohq/views.py
+++ b/backend/ohq/views.py
@@ -245,6 +245,12 @@ class QuestionSearchView(XLSXFileMixin, generics.ListAPIView):
 
     @property
     def paginator(self):
+        """
+        Removes pagination from the XLSX Render so that TAs can download a full list
+        of asked questions.
+        https://www.django-rest-framework.org/api-guide/renderers/#varying-behaviour-by-media-type
+        """
+
         if isinstance(self.request.accepted_renderer, XLSXRenderer):  # xlsx download
             self._paginator = None
             return None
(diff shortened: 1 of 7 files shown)


backend/ohq/views.py
```diff
@@ -222,19 +228,28 @@ def list(self, request, *args, **kwargs):
...
 
     def get_queryset(self):
         qs = Question.objects.filter(
             queue__in=Queue.objects.filter(course=self.kwargs["course_pk"])
         ).order_by("time_asked")
         return prefetch(qs, self.serializer_class)
 
+    @property
+    def paginator(self):
+        if isinstance(self.request.accepted_renderer, XLSXRenderer):  # xlsx download
```
davish, line 254: This seems fine, I'd just leave a more descriptive comment -- maybe link to documentation on `accepted_renderer` and how it's set based on the request?

davish (approved):
Backend changes seem fine. Just more docs would be good.

backend/ohq/views.py
ArmaanT: Makes sense. Added

davish: Do we want to be slightly more general? `if self.request.accepted_handler.format == 'xlsx'`? I don't think in practice there's a difference, but it'd be more "forwards compatible" in case we change something.

PR #36, opened 2020-09-22
Websocket Integration (2nd try)

eyingxuan:
Cherry picked old commits after revert

### DIFF
--- a/frontend/hooks/data-fetching/course.ts
+++ b/frontend/hooks/data-fetching/course.ts
@@ -1,4 +1,11 @@
+import { useEffect } from "react";
 import useSWR, { mutate as globalMutate } from "swr";
+import {
+    useRealtimeResourceList,
+    useRealtimeResource,
+} from "@pennlabs/rest-live-hooks";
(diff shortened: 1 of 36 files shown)


frontend/hooks/data-fetching/course.ts
```diff
@@ -96,50 +104,114 @@ export async function getSemesters(): Promise<Semester[]> {
...
-        (id) => `/courses/${courseId}/queues/${queueId}/last/${id}/`,
-        [],
-        { refreshInterval }
+        (id) => `/courses/${courseId}/queues/${queueId}/last/${id}/`
     );
 
+    const stringified = JSON.stringify(qdata);
+
+    // this revalidates the last question query whenever there is a websocket update
+    useEffect(() => {
```
eyingxuan, line 208: mutate is stale in this effect, this is fine because we only use it for revalidation, but it is bad style

pawalt:
Can we get some auth between django and redis? Even if it's just a simple user/password that's part of the REDIS_URL, I think that'd make me a lot more comfortable.

PR #37, opened 2020-09-26
Notifications

gautamr6:
- Abstracted notification sound hook
- Instructors and students can now toggle notifications
[ch2433] [ch2423]


frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -47,14 +47,7 @@ const Queue = (props: QueueProps) => {
...
 
-    const [player, setPlayer] = useState<UIfx | undefined>();
-    useEffect(() => {
-        setPlayer(
-            new UIfx(bellAudio, {
-                throttleMs: 100,
-            })
-        );
-    }, []);
+    // const play = usePlayer(bellAudio);
```
eyingxuan: Remember to remove this comment!

frontend/pages/courses/[course]/index.tsx
```diff
@@ -34,24 +34,26 @@ const QueuePage = (props: QueuePageProps) => {
                 <CourseWrapper
                     course={course}
                     leadership={leadership}
-                    render={(staff: boolean) => {
+                    render={(staff: boolean, play: () => void) => {
                         return (
-                            <>
+                            <div style={{ marginTop: "-30px" }}>
```
Comment (eyingxuan) [inline frontend/pages/courses/[course]/index.tsx]: Same nit about rem

frontend/hooks/player.ts
```diff
@@ -0,0 +1,25 @@
+import UIfx from "uifx";
+import { useState, useEffect, Dispatch, SetStateAction } from "react";
+
+export function usePlayer(
+    audio: string
+): [boolean, Dispatch<SetStateAction<Boolean>>, () => void] {
+    const [player, setPlayer] = useState<UIfx | undefined>();
```
eyingxuan: I know I used useState to begin with, but could you change this to a ref? It will save one unnecessary render

frontend/components/Course/StudentQueuePage/QuestionCard.tsx
```diff
@@ -47,6 +50,13 @@ const QuestionCard = (props: QuestionCardProps) => {
         });
     };
 
+    useEffect(() => {
+        if (question.status == QuestionStatus.ACTIVE) {
+            play();
+        }
+        // eslint-disable-next-line
+    }, [question.status]);
```
eyingxuan: ESLint is actually right on this one, you shouldn't disable it, and should add "play" to the dependency array. Remember that play() actually depends on notif, so right now if notif changes play() doesn't get updated properly here and might be stale.

frontend/components/Course/StudentQueuePage/QuestionCard.tsx
```diff
@@ -47,6 +50,13 @@ const QuestionCard = (props: QuestionCardProps) => {
         });
     };
 
+    useEffect(() => {
+        if (question.status == QuestionStatus.ACTIVE) {
```
eyingxuan: Use "===" for better style, in fact this should probably be a lint error instead of warning

eyingxuan (changes requested):
In general, great first PR!! Hook is abstracted well and it'll be super easy to add audio players now!

Left some comments, and most of them are just nits about px to rem. After you make changes let's test it one more time and we can merge. 

Congrats on first PR @ OHQ 🎊🥳

eyingxuan:
Can you also add a screenshot to the PR for posteriority?

gautamr6:
notifs on
notifs off

frontend/hooks/player.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function usePlayer(
+    audio: string
+): [
+    boolean,
+    Dispatch<SetStateAction<Boolean>>,
+    MutableRefObject<(() => void) | undefined>
+] {
+    const player = useRef<UIfx>();
+    useEffect(() => {
```
eyingxuan, line 15: This should have an empty dependency array so you don't create a new UIfx every time

frontend/hooks/player.ts
```diff
@@ -0,0 +1,33 @@
...
+        player.current = new UIfx(audio, { throttleMs: 100 });
+    });
+
+    const [notifs, setNotifs] = useState(true);
+
+    const play = useRef<() => void>();
```
eyingxuan: You can instantiate the function by passing it as a parameter to useRef - this makes sure that play.current is never undefined, and you don't have to do the check everywhere else

eyingxuan (approved):
A couple of small things left, but looks good to (squash and) merge!

frontend/hooks/player.ts
eyingxuan: Remember to also change the types elsewhere so you don't have to do play.current checks anymore

PR #38, opened 2020-09-26
Subscribes Queue Serializer

eyingxuan:
The queue model is not currently subscribable :( but ~this PR fixes it~ It does not - questions_asked is a derived field and will not receive subscriptions


eyingxuan:
Closing because this requires more DRL changes

PR #40, opened 2020-09-28
Scale down replicas

eyingxuan:
... and save us money


KeenWill (approved):

ArmaanT (approved):
We're still spending the same amount, but I appreciate the sentiment

PR #44, opened 2020-09-29
Websocket 3rd attempt

eyingxuan:
Why do you not "just work"?

S | -> reverting the S'
S' | -> reverting the S

S

### DIFF
--- a/k8s/values.yaml
+++ b/k8s/values.yaml
@@ -6,7 +6,7 @@ applications:
     image: pennlabs/office-hours-queue-backend
     secret: office-hours-queue
     cmd: ["/usr/local/bin/asgi-run"]
-    replicas: 3
+    replicas: 1
     ingress:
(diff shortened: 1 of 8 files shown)


k8s/values.yaml
```diff
@@ -6,7 +6,7 @@ applications:
     image: pennlabs/office-hours-queue-backend
     secret: office-hours-queue
     cmd: ["/usr/local/bin/asgi-run"]
-    replicas: 3
+    replicas: 1
```
pawalt: Let's keep this at 2. Only reason is that in a deploy, it's nice to have spare compute so you don't have all of your websocket connections getting bounced at once. This is less of a big deal for standard http apps, but it becomes a big deal for websockets since disruptions are pretty unfriendly.

ArmaanT (approved):
v3 of v3

PR #48, opened 2020-10-11
Student leave course button

conswang:
[ch2365](https://app.clubhouse.io/pennlabs/story/2365/leaving-a-course)

Changes Made:
-  add leave student course modal
- add dropdown to course cards for leaving a course
- refactor props to pass down user memberships instead of courses (a user membership consists of id, course, and kind), so we can get the membership id to make a delete request to destroy that membership


frontend/hooks/data-fetching/dashboard.ts
```diff
@@ -19,6 +19,22 @@ export async function joinCourse(courseId: string): Promise<void> {
     }
 }
 
+export async function leaveCourse(
+    courseId: string,
+    membershipId: string
+): Promise<void> {
+    const res = await doApiRequest(
+        `/courses/${courseId}/members/${membershipId}`,
```
ArmaanT: I haven't looked at the rest of this, but all API requests should have a trailing slash

frontend/components/Home/Dashboard/Modals/ModalLeaveStudentCourse.tsx
```diff
@@ -0,0 +1,39 @@
...
+    leaveMembership?: UserMembership;
+    closeFunc: () => void;
+    mutate: mutateFunction<UserMembership[]>;
+}
+
+const ModalLeaveStudentCourse = (props: ModalLeaveStudentCourseProps) => {
+    const { open, leaveMembership, closeFunc, mutate } = props;
+
+    const leaveFunc = async () => {
+        await leaveCourse(
```
eyingxuan: leaveCourse could throw an error here - consider handling it like ModalAddInstructorCourse.tsx

eyingxuan (changes requested):
Looks great mostly! Small thing left before we merge.

frontend/components/Home/Dashboard/StudentCourses.tsx
```diff
@@ -44,11 +76,27 @@ const StudentCourses = (props: StudentCoursesProps) => {
...
+            </Snackbar>
+
+            <Snackbar
+                open={addToastOpen}
                 autoHideDuration={2000}
-                onClose={() => setSuccess(false)}
+                onClose={() => setAddToastOpen(false)}
             >
-                <Alert severity="success" onClose={() => setSuccess(false)}>
+                <Alert
```
eyingxuan, line 81: Instead of using 2 Alert components, you can share them using the pattern 
https://github.com/pennlabs/office-hours-queue/blob/c1e8facca8bff22bf37cdefe8c7d2481ad77fb85/frontend/components/Home/AccountSettings/AccountForm.tsx
here!

eyingxuan (changes requested):

frontend/components/Home/Dashboard/StudentCourses.tsx
conswang: oh wait my bad, I didn't see what you linked to here! Looks a lot cleaner than what I pushed just now... I'll do this instead (. ❛ ᴗ ❛.)

frontend/components/Home/Dashboard/Cards/CourseCard.tsx
```diff
@@ -34,10 +45,31 @@ const CourseCard = (props: CourseCardProps) => {
                             style={{
                                 whiteSpace: "nowrap",
                                 textOverflow: "ellipsis",
-                                overflow: "hidden",
                             }}
                         >
                             {`${course.department} ${course.courseCode}`}
+                            {membership.kind === Kind.STUDENT ? (
```
eyingxuan, line 51: Can replace the ternary operator with membership.kind === Kind.STUDENT &&

eyingxuan: Can replace the ternary operator with membership.kind === Kind.STUDENT &&

frontend/components/Home/Dashboard/StudentCourses.tsx
```diff
@@ -5,36 +5,64 @@ import Alert from "@material-ui/lab/Alert";
...
+
+    const [openLeave, setOpenLeave] = useState(false);
+    const [leaveMembership, setLeaveMembership] = useState<
+        UserMembership | undefined
+    >(undefined);
 
     return (
         <>
             <Grid.Row padded="true">
+                <ModalLeaveStudentCourse
```
eyingxuan: Would it be better if you wrap this component in a check of whether leaveMembership is undefined? Then you would be able to guarantee that leaveMembership is not undefined in the modal (and you'd change the prop types there to reflect that). The advantage is that the LeaveMembershipModal is easier to read since it can be confusing as to why the membership itself is undefined.

frontend/components/Home/Dashboard/Cards/CourseCard.tsx
```diff
@@ -1,19 +1,30 @@
...
-import { Segment, Header } from "semantic-ui-react";
+import { Segment, Header, Dropdown, Icon } from "semantic-ui-react";
 import Link from "next/link";
-import { Course } from "../../../../types";
+import { UserMembership, Kind } from "../../../../types";
 
 interface CourseCardProps {
-    course: Course;
+    membership: UserMembership;
+    setOpenLeave?: (open: boolean) => void;
```
eyingxuan, line 8: More of a side comment - the types as is make it so that it is possible for a student course card to not have setOpenLeave and setLeaveMembership, which is undesirable since we want to enforce that those functions exist if the membership is a student. 

However, I'm actually not sure how to enforce this with types cleanly. If you happen to want to attempt a nice type exercise feel free to think about how this can be fixed, but it's not the most important thing in the world. If you want to leave it as is and get this merged go for it!

eyingxuan (approved):
Small changes left, then you're good to go!

frontend/components/Home/Dashboard/StudentCourses.tsx
conswang: Makes sense, then can get rid of all the question marks!

PR #49, opened 2020-10-16
Tags

ArmaanT:
This PR adds the ability for Head TAs+ to define tags for a course and students have the ability to tag their questions with zero or more of those tags.

~No frontend work has been on in this PR yet~


judtinzhang (approved):

kevc528 (commented):
LGTM, but do we know how to fix the failed checks?

frontend/hooks/data-fetching/course.ts
```diff
@@ -30,6 +31,9 @@ import {
 export const useCourse = (courseId: number, initialCourse: Course) =>
     useResource(`/courses/${courseId}/`, initialCourse);
 
+export const useTags = (courseId: number, initialTags: Tag[]) =>
```
eyingxuan: Saw you attempted to use the resource hooks but ended up writing most of the API calls yourself below. 

You are definitely on the right track, but was using the wrong type of hook. What you're looking for is the useResourceList hook (https://github.com/pennlabs/frontend-core/blob/master/packages/rest-hooks/src/useResourceList.ts source code here). Basically, the hook provides you with tag data (which replaces getTags) and you can delete tags using the `mutate()` function by feeding in the id. Note that even though the routes for deleting a specific tag and getting all tags is slightly different, the 2nd parameter in the resource list hook allows you to specify the route of a specific tag item. A good reference for how to use this hook is useQuestions (which is also in this file). Unfortunately you still need createTag, similar to how we still have createQuestion. 

This should save some boilerplate with useEffects and keeping state in the components.

frontend/components/Course/StudentQueuePage/QuestionForm.tsx
```diff
@@ -51,15 +54,48 @@ const QuestionForm = (props: QuestionFormProps) => {
...
-    // const getDropdownOptions = (tags) => {
-    //     return tags.map((tag) => {
-    //         return {
-    //             key: tag,
-    //             value: tag,
-    //             text: tag,
-    //         };
-    //     });
-    // };
+    useEffect(() => {
```
eyingxuan: Shouldn't we be able to share the tags data with EditQuestionModal? Perhaps but pulling it up to the QuestionCard component. 

Boilerplate can also be saved with resource hook as described below.

frontend/components/Course/InstructorQueuePage/CreateQueue/CreateQueue.tsx
```diff
@@ -82,15 +82,15 @@ const CreateQueue = (props: CreateQueueProps) => {
...
                                 components={{ DropdownIndicator: null }}
-                                inputValue={tagsInputValue}
+                                // inputValue={tagsInputValue}
                                 isClearable
                                 isMulti
                                 menuIsOpen={false}
-                                onChange={handleTagChange}
-                                onInputChange={handleTagsInputChange}
-                                onKeyDown={handleTagsKeyDown}
+                                // onChange={handleTagChange}
```
eyingxuan: This block of commented code should be removed right?

frontend/components/Course/StudentQueuePage/EditQuestionModal.tsx
```diff
@@ -48,15 +59,50 @@ const EditQuestionModal = (props: EditQuestionModalProps) => {
...
-    //             value: tag,
-    //             text: tag,
-    //         };
-    //     });
-    // };
+    useEffect(() => {
+        const fetchData = async () => {
+            const loadedTags: Tag[] = await getTags(course.id);
+
+            setTagOptions(
```
eyingxuan: See comments in QuestionForm.

frontend/components/Course/CourseSettings/CourseForm.tsx
```diff
@@ -138,6 +168,76 @@ const CourseForm = (props: CourseFormProps) => {
...
+                })
+            );
+        } else if (event.action === "clear") {
+            setTagLabels([]);
+        } else if (event.action === "select-option") {
+            handleCreateTag(event.option.label);
+        }
+    };
+
+    useEffect(() => {
```
eyingxuan: This might not work as intended - consider the situation where some other input (say course name) was changed, and a tag was added and removed. But, the save button will still be disabled. This logic belongs to the isValid function.

frontend/components/Course/CourseSettings/CourseForm.tsx
```diff
@@ -138,6 +168,76 @@ const CourseForm = (props: CourseFormProps) => {
             });
     };
 
+    useEffect(() => {
```
eyingxuan: I don't think this is actually getting called when you expect it to be - tags information are not part of the course object. You will have to probably move this logic into onSubmit. 

A side effect of this - I think tags are pretty buggy on the course settings page now. Try interacting with the list after saving once.

eyingxuan (changes requested):
In general, a good first attempt at the PR! Left a couple of comments here and there. Let me know if you have any questions!

ArmaanT:
@eyingxuan and @kevc528 looks like the migrations are a bit off in this PR (duplicate migration 0005). Once this PR has the final approval, just ping me and I'll reset the migrations properly.
It's just a matter of making sure the branch is up to date with master, then just deleting the new migration files and running makemigrations once again if you want to try it as well @kevc528

eyingxuan:
Oh yikes I saw there was a duplicate migration and ran a command to merge the migrations to create 0006 - is that not the right way to do things?

ArmaanT:
Yeah, what happened was we had two parallel PRs that added a 0005 migration (this and the notes one). From here it's a bit of a personal preference on how to fix things, but since the other PR was merged first, I think the easiest thing to do is to just rm all the generated migrations in this PR and then rerun makemigration so that we don't make any modifications to the existing migrations, but then we just generate a single 0006 migration instead of two files to keep track of

eyingxuan:
So it seems like tags information don't get sent over websocket updates - will Davis' latest DRL update patch this problem? @ArmaanT

PR #63, opened 2020-11-06
Populate database

judtinzhang:
Created the populate.py file to populate databases with courses, queues, questions, members, etc.

### DIFF
--- a/backend/ohq/management/commands/populate.py
+++ b/backend/ohq/management/commands/populate.py
@@ -6,7 +6,8 @@ from django.core.files.base import ContentFile
 from django.core.management.base import BaseCommand, CommandError
 from django.utils import timezone
 
-from ohq.models import Course, Membership, MembershipInvite, Question, Queue, Semester, Profile
+from ohq.models import Course, Membership, MembershipInvite, Profile, Question, Queue, Semester
+
 
 now = timezone.now()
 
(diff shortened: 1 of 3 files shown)


backend/ohq/management/commands/populate.py
```diff
@@ -0,0 +1,458 @@
...
+]
+
+
+class Command(BaseCommand):
+    help = "Populates the development environment with dummy data."
+
+    def handle(self, *args, **kwargs):
+
+        # not sure if this line is neccessary / works
+        if Course.objects.filter(course_title="Example Queues for OHQ").exists():
```
eyingxuan: You should be able to use some env var (I think settings.DEBUG?) to check whether you are in production isntead of relying on a specific course

backend/ohq/management/commands/populate.py
```diff
@@ -0,0 +1,458 @@
...
+            raise CommandError("You probably do not want to run this script in production!")
+
+        # ask what "last active" means for membership, should we add that in for ying
```
eyingxuan: Last active is used to determine whether a TA is online - would it be possible to fix everyone to have a last active of at least 1 minute before time.now()? That's the threshold for determining if a TA is online and it's to make sure we don't run into weird snapshot issues when some runs are faster than others.

eyingxuan (approved):
Two small changes (given my limited expertise) - data itself looks great! Thanks for the good work :)

PR #70, opened 2020-12-28
Queue related analytics

kevc528:
This PR allows students and instructors to see analytics related to queues

* Heatmap statistics
    * Average wait-time
    * Questions per instructor
* Card statistics
    * Course-wide (by week)
        * Average wait-time
        * Number of questions answered per week
        * Students helped per week
        * Average time spent helping students
* List statistics
    * Day with highest wait-time

_Other analytics (personal statistics and statistics related to queue size) will be in a separate PR. Hopefully this way frontend can start working as well._

No frontend work has been done for this PR yet


conswang:
Heatmaps for average number of questions (data available for each queue):
image

Dropdown nav to pick queue to see stats for:
image

Issues
- chart is not responsive
- stats only available for non-archived queues

eyingxuan (changes requested):
Minor things left for frontend side

frontend/components/Course/Analytics/Heatmaps/Heatmap.tsx
```diff
@@ -0,0 +1,74 @@
+import dynamic from "next/dynamic";
+import React from "react";
+import { HeatmapSeries } from "../../../../types";
+
+interface HeatmapProps {
+    series: HeatmapSeries[];
+    chartTitle: string;
+}
+
+const Chart = dynamic(() => import("react-apexcharts"), { ssr: false });
```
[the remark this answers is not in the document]
conswang: Not that familiar with Next js dynamic importing, but it seems like react apex charts doesn't work on the server (get "window is not defined" at apexcharts/dist/apexcharts.common.js if importing it normally). I think it might be an import error since not even putting a window !== undefined check around the chart component doesn't work. I saw "when the module includes a library that only works in the browser" was one of the use cases for dynamic imports

Let me know if there's something better though!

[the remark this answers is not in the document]
eyingxuan: Nope, this sounds like the perfect use case for dynamic imports - just leave a small comment in code for future purposes.

frontend/types.tsx
```diff
@@ -160,3 +160,38 @@ export interface CoursePageProps {
...
+    AVG_WAIT = "AVG_WAIT",
+    NUM_ANSWERED = "NUM_ANSWERED",
+    STUDENTS_HELPED = "STUDENTS_HELPED",
+    AVG_TIME_HELPING = "AVG_TIME_HELPING",
+    LIST_WAIT = "LIST_WAIT_TIME_DAYS",
+}
+
+interface AnalyticsDataElem {
+    metric: Metric;
+    day: number;
```
ArmaanT, line 176: According to the django model, `day`, `hour`, and `date` could be undefined. Those fields may have to be optional here

backend/ohq/models.py
```diff
@@ -268,6 +269,89 @@ class Question(models.Model):
...
+        (METRIC_HEATMAP_QUESTIONS_PER_TA, "Questions per TA heatmap"),
+        (METRIC_AVG_WAIT, "Average wait-time"),
+        (METRIC_NUM_ANSWERED, "Number of questions answered per week"),
+        (METRIC_STUDENTS_HELPED, "Students helped per week"),
+        (METRIC_AVG_TIME_HELPING, "Average time helping students"),
+        (METRIC_LIST_WAIT_TIME_DAYS, "List of wait times per day"),
+    ]
+
+    # for specific days during the week - used for heatmap and graphs where day is x-axis
+    DAY_MONDAY = 0
```
ArmaanT: Maybe we should change Sunday to the first day of the week? I think having Monday first could cause some confusion

frontend/types.tsx
```diff
@@ -160,3 +160,38 @@ export interface CoursePageProps {
...
+
+export type AnalyticsData = AnalyticsDataElem[];
+
+export interface HeatmapSeries {
+    name: string;
+    data: { x: string; y: number }[];
+}
+
+export enum DayOfWeek {
+    "Monday" = 0,
```
ArmaanT: Ditto on the Monday comment

backend/ohq/management/commands/avg_queue_wait.py
```diff
@@ -0,0 +1,24 @@
+from django.db.models import Avg, F
```
ArmaanT: In general, I think it makes sense to split out the logic of a command into a different file (just so it's possible to call that function directly instead of being forced to use the command). Do you think it makes sense to move the logic for all of these into a `ohq/statistics.py` or something like that?

ArmaanT (changes requested):
I've got a couple of questions, but this is looking really good!

backend/ohq/views.py
```diff
@@ -551,6 +554,24 @@ def post(self, request, course_pk, format=None):
...
+    and singleton for card statistics.
+    """
+
+    serializer_class = QueueStatisticSerializer
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = QueueStatisticFilter
+    permission_classes = [QueueStatisticPermission | IsSuperuser]
+
+    def get_queryset(self):
+        # might need to change qs if students shouldn't be able to see all queue statistics
```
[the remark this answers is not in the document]
eyingxuan: I think it's fine for queue statistics to be visible for students, since they're very macro averages and does not contain sensitive info. (Although I forgot that we haven't included analytics for the student page yet - we will get to this in another PR @conswang )

frontend/types.tsx
eyingxuan: Ah good catch - some metrics have day/hour/date and some don't, but this is known ahead of time so @conswang can you turn this into a discriminated union with metric being the discriminator?

backend/ohq/models.py
eyingxuan: Should we also just follow django ORM conventions by using 1-7?

frontend/types.tsx
conswang: Sure, so something like AnalyticsDataElem = HeatmapDataElem | AverageDataElem | ...  where each type may or may not have date/hour/etc, right

Also, I've just noticed that we seem to be returning all the fields from the backend for most? metrics, though they may be null (eg. `date` for heatmaps, `day` and `hour` for averages). Is it safe to assume that the fields that are null are the ones we don't need for a particular metric?

image

eyingxuan: Yeap! And as for backend response, yes we can assume that for heatmaps there won't be a date because the values are averages, and for averages, they are day-to-day statistics that give you data to plot graphs, but don't have hour granularity.

backend/ohq/models.py
kevc528: Yep will reorder the days to be Sunday = 1, Monday = 2, etc. The frontend will also need to be updated then.

backend/ohq/models.py
```diff
@@ -268,6 +269,89 @@ class Question(models.Model):
     tags = models.ManyToManyField(Tag, blank=True)
 
 
+class QueueStatistic(models.Model):
```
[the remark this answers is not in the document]
kevc528: Yep will add that

backend/ohq/management/commands/avg_queue_wait.py
kevc528: Yep will do! After the changes, the commands that need to be run are queue_weekly_stat (groups weekly stats), queue_heatmap_stat (groups heatmap stats), and wait_time_days.

frontend/types.tsx
conswang: Fixed it I think! I actually didn't put the union type since we wouldn't really be using it for anything, we always want either an array of heatmap data or an array of another kind of averages data (with metric, value, and date)

eyingxuan (approved):
Everything looks good! I will merge this in and run the management commands midnight tonight to prevent db load from overwhelming prod.

PR #71, opened 2021-01-01
Upgrade Django REST Live integration

davish:
This PR moves us to DRL 1.0, fully integrated into REST Framework


eyingxuan (approved):
we be mergin at midnight

PR #72, opened 2021-01-01
Data fetching refactor

conswang:
Refactor all hooks to use functions from @pennlabs/rest-hooks package instead of resources.ts


frontend/hooks/data-fetching/course.ts
```diff
@@ -122,33 +111,28 @@ export async function sendMassInvites(
...
 export const useQueues = (courseId: number, initialData: Queue[]) =>
-    useResourceListNew<Queue>(
+    useResourceList<Queue>(
         `/api/courses/${courseId}/queues/`,
         (id) => `/api/courses/${courseId}/queues/${id}/`,
         {
             initialData,
-            fetcher: newResourceFetcher,
             refreshInterval: QUEUE_STATUS_POLL_INTERVAL,
             refreshWhenHidden: true,
```
eyingxuan: You can also get rid of these refreshWhenHidden fields, since they are configured in the global SWRConfig. They were here before because of our peer dep issue resulting in SWRConfig not being able to configure rest-hooks from last time.

eyingxuan (approved):
LGTM, minor change needed! merge when backend tests pass

PR #78, opened 2021-01-19
Queue rate limiting

kevc528:
This PR gives the option for course staff to rate limit queues, which can prevent students from asking too many questions in a short period of time when the queue is long (i.e. a student can only ask X questions in Y minutes if the queue is longer than Z)

TODOS (so close)
- [x] Modal/snackbar error message about rate-limiting restrictions

Screen Shot 2021-01-20 at 12 20 39 PM
Screen Shot 2021-01-20 at 12 20 45 PM
Screen Shot 2021-01-20 at 12 21 16 PM
Screen Shot 2021-01-20 at 12 21 23 PM
Screen Shot 2021-01-20 at 12 22 12 PM


backend/tests/ohq/test_views.py
```diff
@@ -93,3 +93,76 @@ def test_valid_emails(self):
         content = json.loads(response.content)
         self.assertEqual(1, content["membersAdded"])
         self.assertEqual(1, content["invitesSent"])
+
+
+class QuestionViewTestCase(TestCase):
```
ArmaanT, line 98: I know that I haven't done this, but could you add a docstring here just clearly explain the test case? It should make things a little clearer than just reading the tests

backend/ohq/views.py
```diff
@@ -237,6 +237,45 @@ def list(self, request, *args, **kwargs):
...
         return super().list(request, *args, **kwargs)
 
+    def create(self, request, *args, **kwargs):
+        """
+        Create a new question and check if it follows the rate limit
+        """
+
+        queue = Queue.objects.get(id=self.kwargs["queue_pk"])
+        if (
+            queue.rate_limit_length is not None
```
ArmaanT: I don't know how the frontend will eventually try to change the rate limit, but we my have to do a different check than not None (Maybe a `!= -1` or something. ee can figure that out as the frontend gets built for this

ArmaanT (changes requested):
A few minor points. This is looking really good though!

backend/ohq/views.py
eyingxuan: Since the rate limit field triple comes together, would be the easiest if we have another `rate_limited_enabled` boolean field.

PR #79, opened 2021-01-22
Allow faculty members to create courses

ArmaanT:
This PR allows faculty members to create courses even if they currently don't currently hold a leadership role in an existing ohq course.


frontend/components/Home/Dashboard/Dashboard.tsx
```diff
@@ -38,14 +38,17 @@ const Dashboard = () => {
         });
     };
 
+    const isFaculty = initalUser?.groups.includes("platform_faculty") || false;
```
ArmaanT: @eyingxuan this can't be the best way of doing this, right? The problem is, if I don't add `|| false` the type of `isFaculty` is `boolean | undefined`

eyingxuan: That's because the ? operator propagates the undefined - I think the best way to do this is to add an if statement up top that throws an error if initialUser is undefined (since it should always be defined when this component is loaded). That way TS will be able to determine that initialUser is always non-undefined.

eyingxuan: Also, seems to be a typo with initialUser lol

eyingxuan (approved):
Looks good outside of minor changes to frontend

kevc528 (approved):
Backend looks good to me!

frontend/components/Home/Dashboard/Dashboard.tsx
ArmaanT: haha, I noticed that, but was too lazy to fix it, I guess I'll do it the right way

eyingxuan (approved):
Good to yeet

PR #80, opened 2021-01-22
Kraken

ArmaanT:
This PR migrates OHQ to kraken. Eventually `backend/pyproject.toml` should be deleted so that way black reverts back to it's default line length of 88 characters (but that needs to happen when there are no active backend PRs)


eyingxuan (approved):
lgtm!

PR #87, opened 2021-02-06
Fix question per TA heatmap calculation

kevc528:


backend/ohq/statistics.py
```diff
@@ -92,20 +93,22 @@ def calculate_num_students_helped(queue, prev_sunday, coming_saturday):
...
 
 def calculate_questions_per_ta_heatmap(queue, weekday, hour):
-    interval_questions = Question.objects.filter(
-        queue=queue, time_asked__week_day=weekday, time_asked__hour=hour
+    interval_stats = (
+        Question.objects.filter(queue=queue, time_asked__week_day=weekday, time_asked__hour=hour)
+        .annotate(date=TruncDate("time_asked"))
+        .values("date")
+        .annotate(
+            q_per_ta=1.0 * Count("date", distinct=False) / Count("responded_to_by", distinct=True)
```
eyingxuan: Discovered another potential bug on my test db - responded_to_by could be 0 if there happens to be a slot with one question asked by never responded by anyone which causes a divide-by-zero error. 

Using a case/when/default should work here. 
https://docs.djangoproject.com/en/3.1/ref/models/conditional-expressions/

backend/tests/ohq/test_commands.py
```diff
@@ -596,7 +593,10 @@ def test_questions_per_ta_computation(self):
         yesterday = timezone.datetime.today().date() - timezone.timedelta(days=1)
         yesterday_weekday = yesterday.weekday()
 
-        expected_8 = self.ta_1_questions_8 / self.num_tas_8
+        expected_8 = (
+            self.ta_1_questions_8_week_1 / self.num_tas_8
```
eyingxuan, line 611: Can we modify this test case to catch possible regression where self.num_tas_8 > 1?

backend/ohq/statistics.py
eyingxuan: Can we also add a test case to catch this regression?

eyingxuan (changes requested):
There's a divide-by-zero error that I just found, but other than that we should be good to merge!

eyingxuan (approved):
lgtm!

PR #94, opened 2021-02-20
Auto signup

conswang:
ch 2396
Allow direct signup from course link if the course is not invite-only
Redirect from course link to dashboard page with signup modal
image


frontend/pages/index.tsx
```diff
@@ -1,14 +1,36 @@
...
+import { Toast } from "../types";
 
 const LandingPage = () => {
     const { user } = useContext(AuthUserContext);
+    const { query } = useRouter();
+
+    const [toast, setToast] = useState<Toast>({ message: "", success: true });
+    const [toastOpen, setToastOpen] = useState(false);
+
+    const [signUpCourse] = useCourse(Number(query.signup), undefined);
```
eyingxuan: This approach results in querying for NaN when there are no query parameters!

eyingxuan (changes requested):
A good attempt! Some comments:

The approach taken will look for the query parameter on the client-side, which is strictly speaking fine, but your implementation makes it so that when there are no query parameters, the hook still runs and that results in an invalid query. A strategy to fix this might be to move the hook into the Modal component, and render the component when there is *any* query parameter, then do validation in the modal component.

Since we are using Next, there is actually an even better approach - you can do the data fetching logic in getInitialProps in `index.tsx`, so the client won't actually have to do any GET requests/do validation on the client side, saving you the extra `useCourse` hook.

eyingxuan (approved):
lgtm! merge after cherry picking @kevc528's fix for backend tests/pull from master if he pushes the fix there

PR #95, opened 2021-03-06
Fix course retrieve permission

kevc528:
This PR changes the backend permission so a user can retrieve any course (including courses they aren't a member of). Note that the queryset in the CourseViewSet will make sure that invite only courses that a user isn't a member of can't be accessed.


eyingxuan (approved):
lgtm

PR #98, opened 2021-03-20
work in progress notifications

lhuang19:


eyingxuan:
There seems to be some linting issues - can you fix those before review? Would like to make sure there aren't any type errors as well which might affect how you've implemented some things

frontend/hooks/player.ts
```diff
@@ -10,21 +10,38 @@ import {
 
 export function usePlayer(
     audio: string
-): [boolean, Dispatch<SetStateAction<Boolean>>, MutableRefObject<() => void>] {
+): [
+    boolean,
+    Dispatch<SetStateAction<Boolean>>,
+    MutableRefObject<(string) => void>
```
eyingxuan: Seeing as this type is commonly passed as props, consider adding a type alias and use it so we can easily change the type later on if needed?

frontend/components/Course/StudentQueuePage/StudentQueues.tsx
```diff
@@ -10,7 +10,7 @@ interface StudentQueuesProps {
     course: Course;
     queueMutate: mutateResourceListFunction<Queue>;
     questionmap: QuestionMap;
-    play: MutableRefObject<() => void>;
+    play: MutableRefObject<(string) => void>;
```
eyingxuan: Use type alias here

frontend/components/Course/StudentQueuePage/StudentQueuePage.tsx
```diff
@@ -15,7 +15,7 @@ interface StudentQueuePageProps {
     course: Course;
     queues: Queue[];
     questionmap: QuestionMap;
-    play: MutableRefObject<() => void>;
+    play: MutableRefObject<(string) => void>;
```
eyingxuan: Use type alias here too!

frontend/components/Course/CourseWrapper.tsx
```diff
@@ -65,47 +69,51 @@ const CourseWrapper = ({ render, ...props }: CourseProps) => {
...
-                                            fontSize: "1.29rem",
-                                            fontFamily: "Lato",
-                                            color: "#666666",
+                                            float: "right",
+                                            paddingTop: "0.71rem",
                                         }}
                                     >
-                                        Notifications are{" "}
-                                        {notifs ? "ON" : "OFF"}
+                                        <div
```
eyingxuan: I think we talked about this, and on second thought perhaps we should add a small info indicator here that displays on hover and lets the user know that they should enable notification permissions if possible. 

I think we can even use https://developer.mozilla.org/en-US/docs/Web/API/Notification/permission to tell whether we have permissions, but I'm actually not super sure if there's a great way to re-render the component when it changes, so it *might* be stale so we don't have to do this. If you know how to do this that'll be awesome though

frontend/hooks/player.ts
```diff
@@ -10,21 +10,38 @@ import {
...
     useEffect(() => {
         player.current = new UIfx(audio, { throttleMs: 100 });
     }, [audio]);
 
-    const [notifs, setNotifs] = useState(true);
+    const [notifs, setNotifs] = useState(false);
 
-    const playFunc = () => {
+    useEffect(() => {
+        if (localStorage != null) {
```
eyingxuan, line 23: Let's make the default (no localstorage with notifs === false) be notifications true - or else everyone would have to manually opt in to notifications

frontend/components/Course/InstructorQueuePage/QuestionCard.tsx
```diff
@@ -39,6 +35,9 @@ const QuestionCard = (props: QuestionCardProps) => {
 
     const [open, setOpen] = useState(false);
     const [messageModalOpen, setMessageModalOpen] = useState(false);
+
+    const [lastNotif, setLastNotif] = useState(notifs);
```
eyingxuan, line 40: Add a quick comment on what this does

eyingxuan (approved):
Looks good! Couple of small comments to fix but ready to squash and merge after - congrats!

PR #102, opened 2021-03-28
[WIP] Show Active Staff Information

kevc528:
This PR allows queues to provide information about active staff through an added action under MembershipViewSet to get all staff that are active for a course.

### DIFF
--- a/backend/ohq/views.py
+++ b/backend/ohq/views.py
@@ -376,24 +376,11 @@ class QueueViewSet(viewsets.ModelViewSet):
         questions_active = questions.filter(status=Question.STATUS_ACTIVE)
         questions_asked = questions.filter(status=Question.STATUS_ASKED)
 
-        time_threshold = timezone.now() - timedelta(minutes=1)
-        staff_active = (
-            Membership.objects.filter(
-                Q(course=OuterRef("course__pk"))
-                & ~Q(kind=Membership.KIND_STUDENT)
-                & Q(last_active__gt=time_threshold)
-            )
-            .order_by()
-            .values("course")
-            .annotate(count=Count("*", output_field=FloatField()),)
-            .values("count")
-        )
         qs = (
             Queue.objects.filter(course=self.kwargs["course_pk"], archived=False)
             .annotate(
                 questions_active=Subquery(questions_active[:1], output_field=IntegerField()),
                 questions_asked=Subquery(questions_asked[:1]),
-                staff_active=Subquery(staff_active[:1]),
             )
             .order_by("id")
         )
@@ -483,6 +470,23 @@ class MembershipViewSet(viewsets.ModelViewSet):
                 | Q(user=self.request.user)
             )
         return prefetch(qs, self.serializer_class)
+
+    @action(detail=False, url_path="staffActive")
+    def staff_active(self, request, course_pk):
+        """
+        Get's the active staff in a course
+        """
+    
+        time_threshold = timezone.now() - timedelta(minutes=1)
+
+        staff_active = Membership.objects.filter(
+            Q(course=course_pk)
+            & ~Q(kind=Membership.KIND_STUDENT)
+            & Q(last_active__gt=time_threshold)
+        )
+
+        serializer = self.get_serializer(staff_active, many=True)
+        return Response(serializer.data)
 
 
 class MembershipInviteViewSet(viewsets.ModelViewSet):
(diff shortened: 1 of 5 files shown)


backend/ohq/views.py
```diff
@@ -484,6 +471,23 @@ def get_queryset(self):
...
+        time_threshold = timezone.now() - timedelta(minutes=1)
+
+        staff_active = Membership.objects.filter(
+            Q(course=course_pk)
+            & ~Q(kind=Membership.KIND_STUDENT)
+            & Q(last_active__gt=time_threshold)
+        )
+
+        serializer = self.get_serializer(staff_active, many=True)
+        return Response(serializer.data)
```
ArmaanT, line 506: This definitely works (and is arguably more clear than what I'm about to suggest) but I think you could also have this method just `pass` and then in `get_queryset` check if the action is staff_active and only return active TAs that way

ArmaanT (approved):
Backend looks good to me, I have one comment that we can talk about more if you want. I assume we'll keep this PR open for frontend to make the changes they need?

backend/ohq/views.py
ArmaanT: Actually, after @eyingxuan's presentation on WS I think we should do it this way so the action is WS subscribable

kevc528: I think DRL doesn't allow us to do WS subscribes to custom actions at the moment so I instead used a query param to see if we want to get staff_active vs. list all staff for a course

ArmaanT: Hmm, that's unfortunate. @davish is that the case?

davish: Yep. You subscribe to a queryset as it's visible in a viewset's get_queryset method with the serializer defined in get_serializer. There's no real concept of actions beyond list and retrieve, and there's no way for DRL to know what the action does and to trigger a broadcast message.

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -110,11 +121,31 @@ const Queue = (props: QueueProps) => {
...
-                                content={`${
-                                    queue.staffActive || 0
-                                } staff active`}
-                                icon={<Icon name="sync" loading={true} />}
+                            <Popup
+                                content={
+                                    activeStaff && activeStaff.length > 0 ? (
+                                        <List divided verticalAlign="middle">
+                                            {activeStaff.map((staff) => (
+                                                <List.Item basic>
```
eyingxuan, line 129: You're missing the key prop here! (Always check console for any warnings)

frontend/components/Course/StudentQueuePage/StudentQueue.tsx
```diff
@@ -166,11 +176,40 @@ const StudentQueue = (props: StudentQueueProps) => {
...
+                                <Popup
+                                    content={
+                                        activeStaff &&
+                                        activeStaff.length > 0 ? (
+                                            <List
+                                                divided
+                                                verticalAlign="middle"
+                                            >
+                                                {activeStaff.map((staff) => (
+                                                    <List.Item basic>
```
eyingxuan, line 188: Ditto about key prop

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -110,11 +121,31 @@ const Queue = (props: QueueProps) => {
...
                         {queue.active && (
-                            <Label
-                                content={`${
-                                    queue.staffActive || 0
-                                } staff active`}
-                                icon={<Icon name="sync" loading={true} />}
+                            <Popup
+                                content={
+                                    activeStaff && activeStaff.length > 0 ? (
+                                        <List divided verticalAlign="middle">
```
eyingxuan, line 127: I personally like the aesthetic of List without divided - the line seems a bit weird. Were there any issues if this wasn't divided?

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -22,6 +32,7 @@ const Queue = (props: QueueProps) => {
     const {
         courseId,
         queue,
+        activeStaff,
```
eyingxuan, line 35: I don't think we want activeStaff to be queried just on page load - you should write a custom hook (in `hooks/data-fetching`) that calls useResource with this value as the initial value and the activeStaff route as the URL, and set up a refresh interval so we make the API call every 30 seconds. This pattern is pretty common throughout the codebase - let me know if you have any questions about this.

You'll have to do the same thing for StudentQueue, although you should be able to share the hook.

eyingxuan (changes requested):

PR #104, opened 2021-04-05
Analytics convert utc to local hours

conswang:
Handles non-integer UTC offset timezones the default way Javascript Date does (aka round to integer)
Also fixed the hour label getting cut off

image


eyingxuan:
Correct me if I'm wrong, but I think it might be slightly more complicated than this. The bottom left cell should represent Sunday 12AM UTC, but this should be **Friday** 8PM EST, so we still have to shift some cells around. In particular, the 12AM column represents the change in day now

eyingxuan (commented):
Correct me if I'm wrong, but I think it might be slightly more complicated than this. The bottom left cell should represent Sunday 12AM UTC, but this should be Friday 8PM EST, so we still have to shift some cells around. In particular, the 12AM column represents the change in day now

conswang:
> Correct me if I'm wrong, but I think it might be slightly more complicated than this. The bottom left cell should represent Sunday 12AM UTC, but this should be Friday 8PM EST, so we still have to shift some cells around. In particular, the 12AM column represents the change in day now

Ah I originally thought it was okay to leave the hour columns shifted, since the bulk of waking hours are kind of in the middle in UTC time. But I forgot that the days actually overlap, so a tuesday 10 pm cell could be someone else's wednesday 1 am cell rip, thanks for catching

conswang:
Now looks like this: (sunday 12 AM is at bottom left corner)
image

conswang:
Tested the following values with console logs and by changing my computer's time zone settings:
`utcToLocal(1, 5, new Date())` with UTC-4 => `{localDay: 7, localHour: 21}`
`utcToLocal(4, 4, new Date())` with UTC-4 => `{localDay: 4, localHour: 0}`
`utcToLocal(4, 4, new Date())` with UTC-4 => `{localDay: 4, localHour: 0}`
`utcToLocal(6, 18, new Date())` with UTC+10 => `{localDay: 7, localHour: 4}`
`utcToLocal(7, 14, new Date())` with UTC+10 => `{localDay: 1, localHour: 0}`

Day and hour wrap correctly for both positive and negative UTC offsets - let me know if there's another edge case that would be good to check

eyingxuan:
There is one small edge case left - there are some countries (e.g. India) that have time zones that are not whole hour offsets. I think it will probably be easiest to change the labels for these cases to be non whole hour as well.

conswang:
~~Not beautiful, but it works~~
~~Hour does not wrap around correctly for negative UTC offset (aka UTC-3:30 is displayed as UTC-4:30)~~
Ignore the above
image

conswang:
India timezones (UTC+5:30) - UTC Thursday 9 pm becomes local time Friday 2:30 am
image

Updated logic to display AM or PM to be more readable
Checked that other time zones didn't break

eyingxuan (approved):
amazing! let's get this merged

PR #114, opened 2021-04-25
Integration Testing

gautamr6:


eyingxuan:
> Nice! Glad to see this works :) Can you turn this into async/await syntax instead of having nested callbacks? I think it will look cleaner and be easier to read.

Never mind, Cypress chains don't support async/await

PR #100, opened 2021-03-21
Daily queue statistics

m2chan:
Updated queue statistics (average queue wait time, average time helping, number of questions answered, and number of students helped) to pull daily rather than weekly values.

PR #115, opened 2021-04-25
Remove wait time days job

kevc528:
This PR removes references to the wait_time_days job that was removed in #100


eyingxuan (approved):
Merge when CI in master passes

PR #116, opened 2021-04-26
notification wrapper and ui fix

lhuang19:


frontend/hooks/notifcation.ts
```diff
@@ -0,0 +1,33 @@
+export function askNotificationPermissions() {
```
eyingxuan: Can you move this file to the `utils/` folder instead? Not exactly a hook

frontend/components/Course/CourseWrapper.tsx
```diff
@@ -116,7 +117,7 @@ const CourseWrapper = ({ render, ...props }: CourseProps) => {
                                             </p>
                                             {typeof Notification !==
                                                 "undefined" &&
-                                                Notification.permission ==
+                                                Notification.permission ===
```
eyingxuan, line 123: Did you mean to use `getPermissions` here?

frontend/components/Course/CourseWrapper.tsx
```diff
@@ -101,6 +101,7 @@ const CourseWrapper = ({ render, ...props }: CourseProps) => {
                                                     {notifs ? "ON" : "OFF"}
                                                 </div>
                                             }
+                                            size="mini"
                                         >
                                             <p>
                                                 Browser permissions are{" "}
```
eyingxuan, line 110: I would just not display anything regarding permissions if Notifications is not in window.

frontend/hooks/notifcation.ts
```diff
@@ -0,0 +1,33 @@
...
+            console.log("Notifications API not supported on this device");
+        }
+    }
+}
+
+export function checkPermissions() {
+    return localStorage && localStorage.getItem("notifs") === "false";
+}
+
+export function getPermissions() {
```
eyingxuan: Not exactly sure what the return type will be - I think it might be bool | string? Perhaps it makes sense to return undefined in the case where Notifications are not defined, and a boolean for whether you have permissions for the other case. I would split the function to see if Notifications exist into a helper, and call it throughout the functions in this file.

eyingxuan (commented):
Couple of nits to handle before merging

frontend/components/Course/CourseWrapper.tsx
conswang: This should work: https://developer.mozilla.org/en-US/docs/Web/API/Notification/permission

frontend/hooks/notifcation.ts
conswang: Removed this function since it's not actually being called

frontend/components/Course/CourseWrapper.tsx
eyingxuan: Ah, this was referring to the function "getPermissions()" defined below, but I'm fine with it being removed and inlined here

eyingxuan (approved):
Assuming this works locally, code lgtm!

PR #127, opened 2021-08-18
Upgrade Bomb and Auth Flow Refactor 

eyingxuan:
This PR:

- Upgrades us to Next 11, React 17, Typescript 4.3 and Eslint 7
- Refactors authentication flow to make use of getServerSideProps and Redirect support from Next 11


eyingxuan:
Some issues I realized:
- withAuthComponent might actually not be necessary - we can just accept user props for components with transformed getServerSideProps

frontend/components/Course/CourseSidebar.tsx
```diff
@@ -184,6 +184,7 @@ const CourseSidebar = (props: CourseSidebarProps) => {
             >
                 <a
                     href="https://airtable.com/shrIZxIjyAE3gOUSg"
+                    rel="noreferrer"
```
conswang, line 187: Just curious why we added this?

eyingxuan: Sorry I missed this comment - it was a lint warning, and to be honest it should probably be a lint error because it is a security vulnerability. See here for more details: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md

PR #130, opened 2021-08-31
Archive courses for fall 2021- update dashboard welcome message on front end

conswang:
New message:
image

Tested that "__fall_2021_transition_message_shown" is set to true after page renders the first time in local storage
Tested that message disappears after refresh OR clicking x icon

### DIFF
--- a/frontend/constants.ts
+++ b/frontend/constants.ts
@@ -18,3 +18,5 @@ export const SPRING_2021_TRANSITION_MESSAGE_TOKEN =
     "__spring_2021_transition_message_shown";
 export const ANALYTICS_SURVEY_SHOWN_SPRING_2021_END_TOKEN =
     "__analytics_survey_sp_2021_end_shown";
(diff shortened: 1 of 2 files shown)


frontend/constants.ts
```diff
@@ -18,3 +18,5 @@ export const SPRING_2021_TRANSITION_MESSAGE_TOKEN =
     "__spring_2021_transition_message_shown";
 export const ANALYTICS_SURVEY_SHOWN_SPRING_2021_END_TOKEN =
     "__analytics_survey_sp_2021_end_shown";
```
conswang, line 20: Can I delete the above 3 constants?

eyingxuan: I would lean towards not deleting them, just so we don't accidentally choose keys that conflict with these in the future. As these grow larger we might want to move them into a separate file of have a better API to create/access new keys.

eyingxuan (approved):
lgtm!

PR #139, opened 2021-09-19
Upgrading Core Libs

eyingxuan:
Upgrade to Next 11, React 17, Typescript 4, Eslint 7 and make associated lint fixes


frontend/components/Course/CourseWrapper.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { useContext, MutableRefObject } from "react";
+import { useContext, MutableRefObject } from "react";
+import * as React from "react";
```
conswang, line 2: Looked at this component, I don't think we need this import anymore

Commenting so I remember to look at the toggleNotifs function again

conswang (commented):
Yayyy we don't have to import React every time... What did they think we would use, Vue?

Just a few things I want to clarify
- we intentially want to delete the commented out code in MyPieChart.tsx and NewUserForm.tsx and don't want to revisit these components later
- deleting the yarn.lock file?
- what does change to docker file mean

frontend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10.18.0-buster-slim
+FROM node:14-buster-slim
```
conswang, line 1: What are we upgrading here? Trying to learn a bit more about Docker

ArmaanT: That's the version of node. We're currently using 10.18 but this PR would update us to 14!

frontend/next-env.d.ts
```diff
@@ -1,2 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
+/// <reference types="next/image-types/global" />
```
conswang, line 3: Is this change generated from the next 11 upgrade or separate? Looks like it's for importing images, does it come from https://nextjs.org/blog/next-10#built-in-image-component-and-automatic-image-optimization ?

frontend/Dockerfile
conswang: Thanks sounds good :>

frontend/next-env.d.ts
eyingxuan: Yeap, this comes from the Next 11 upgrade! I'm honestly not quite sure what this file does, but it is autogenerated by Next.

frontend/components/Course/CourseWrapper.tsx
eyingxuan: This seems to be used in toggleNotifs, but the parameters are untouched so I removed them - good point!

eyingxuan:
> we intentially want to delete the commented out code in MyPieChart.tsx and NewUserForm.tsx and don't want to revisit these components later

Ah, I didn't actually do anything to them, but the codemod must have decided to remove comments somehow. They were due to be removed though, so I got rid of them (along with NewUserModal).

> deleting the yarn.lock file?

There were just a lot of changes to the yarn.lock file, but it's still there! Diff is just too big to be viewed.

conswang (approved):
letss gooo :>>

PR #149, opened 2021-10-03
[WIP] first attempt at styled components for media query

conswang:


conswang:
See fix/mobile-create-announcement-react-responsive instead

PR #156, opened 2021-10-08
Enable pins for in-person OH

bqle:
This PR allows TAs to restrict to only in-person questions


backend/ohq/models.py
```diff
@@ -198,6 +198,8 @@ class Queue(models.Model):
     question_template = models.TextField(blank=True, default="")
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
     archived = models.BooleanField(default=False)
+    is_pin_enabled = models.BooleanField(default=False)
+    pin = models.CharField(max_length=5, default="", blank=True, null=True)
```
kevc528: Looking at this again, I don't think we actually need the `default=""` here since we added that this field can be null, meaning pin would just be set to null in cases that it isn't provided

kevc528 (changes requested):

backend/ohq/models.py
```diff
@@ -198,6 +198,8 @@ class Queue(models.Model):
     question_template = models.TextField(blank=True, default="")
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
     archived = models.BooleanField(default=False)
+    is_pin_enabled = models.BooleanField(default=False)
```
ArmaanT: It doesn't really matter, but my person preference for this would be `pin_enabled`

backend/ohq/serializers.py
```diff
@@ -164,15 +170,34 @@ def update(self, instance, validated_data):
...
+        if "pin" in validated_data:
+            instance.pin = validated_data["pin"]
 
         instance.save()
         return instance
 
+    def get_pin(self, instance):
+        # only TAs can get pin
+        user = self.context["request"].user
+        membership = Membership.objects.get(course=instance.course, user=user)
```
ArmaanT: I know this should technically work as-is because you need to be in a class to be able to view the queue, but something like `Membership.objects.filter(course=instance.course, user=user).first()` would be better because `.get()` will raise an Exception where filter will just return None

backend/ohq/serializers.py
```diff
@@ -164,15 +170,34 @@ def update(self, instance, validated_data):
         user = self.context["request"].user
         membership = Membership.objects.get(course=instance.course, user=user)
 
+        # generate a random pin when the queue is opened and the queue has pin enabled
+        if "active" in validated_data and validated_data["active"] and instance.is_pin_enabled:
+            validated_data["pin"] = "".join(
```
ArmaanT: We use `get_random_string` elsewhere in OHQ so we should probably stick to the same method for consistency https://github.com/pennlabs/office-hours-queue/blob/fb7aad968f65828f46d42d0bf5f86a26c822cdf5/backend/ohq/serializers.py#L368

PR #160, opened 2021-10-12
Multi-line descriptions and announcements

KeenWill:
Similar to #159, this PR enabled support for multi-line descriptions and announcements.

image


eyingxuan:
Does the announcement message end with a newline? It seems like adding the <p> tag introduces too much margin between the "Posted ..." message and the content.

KeenWill:
> Does the announcement message end with a newline? It seems like adding the
> 
> tag introduces too much margin between the "Posted ..." message and the content.

Resolved in most recent commit

PR #166, opened 2021-10-24
changelog files

lhuang19:


conswang:
Bottom popup and link to changelog in footer
image

Modal version of changelog with diff
image

Page version of changelog with diff
image

Page version of changelog, no diff
image

frontend/components/Changelog/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+    };
+    const [mdLine, setMdLine] = useState<mdLine[]>([initial]);
+    const [showSlider, setShowSlider] = useState(false);
+
+    useEffect(() => {
+        Promise.all([
+            fetch("./changelog.md").then((md) => md.text()),
+            window.localStorage.getItem("changelogsaved") == null
+                ? ""
+                : window.localStorage.getItem("changelogsaved")!,
```
conswang: Don't think we need the ! since there's already a null check

frontend/components/Changelog/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+        ]).then(([readIn, savedMd]) => {
+            if (readIn !== savedMd) setShowSlider(true);
+            const diff = diffLines(savedMd, readIn);
+            const newMd: Array<mdLine> = [];
+            diff.forEach((part) => {
+                let lineColor = "#00000000";
+                if (part.added) {
+                    lineColor = "#7CFC0070";
```
conswang: Maybe assign hex codes to a constant like DIFF_GREEN = "#7CFC0070" so it's easier to tell what colours these are

conswang (approved):

PR #172, opened 2021-11-07
remove notification text on mobile

lhuang19:
Old notification text with icon
image

New just icon
image


conswang (approved):
Looks like mobile notifications aren't playing sound, so aren't actually working - need to look into it some more

But we should still remove the extra text because it's overlapping with the title text

PR #174, opened 2021-11-07
remove min height on queue list

lhuang19:
Old: Queue takes up too much space on mobile
image

New: Queue is collapsed
image


conswang (approved):
nice!

PR #177, opened 2021-11-18
Fix API request error handling

danielxue:
Previous error handling for API requests sometimes threw non-error object exceptions, which were logged in sentry incorrectly. Further, status codes were not being returned, which caused a bug in QuestionForm when checking for error codes. 

Changed so proper error objects always returned, change createQuestion return to number | undefined promise in order to allow for returning error codes, and moved status code handling from catch block to if statements in resolve function.


frontend/components/Course/InstructorQueuePage/CreateQueue/CreateQueue.tsx
```diff
@@ -126,7 +125,6 @@ const CreateQueue = (props: CreateQueueProps) => {
             backFunc();
             successFunc();
         } catch (e) {
-            logException(e);
```
conswang, line 129: Actually, now that the createQueue function handles the error itself and doesn't throw the error, I checked that none of the other functions in this try catch block throw errors.

So we should get rid of the try/catch. This may apply to some of the other components you took the logException(e) out of

frontend/components/Course/StudentQueuePage/QuestionForm.tsx
```diff
@@ -85,24 +84,28 @@ const QuestionForm = (props: QuestionFormProps) => {
...
+            message = "Question successfully added to queue";
+            props.toastFunc(message, null);
+            await props.queueMutate(-1, null);
         }
+        await props.mutate(-1, null);
+    };
+
+    const onSubmit = async () => {
+        setCreatePending(true);
+        await createQuestion(courseId, queue.id, input).then(handleSubmission);
```
conswang: Don't mix .then with await. If you want to unwrap the promise returned by the createQuestion function, then pass it to handleSubmission, do either

const status = await createQuestion(courseId, queue.id, input)
handleSubmission(status)

OR delete await and async

frontend/hooks/data-fetching/course.ts
```diff
@@ -107,10 +108,9 @@ export async function sendMassInvites(
 
     if (!res.ok) {
         const error = await res.json();
-        throw {
-            payload: JSON.stringify(payload),
-            error: JSON.stringify(error),
-        };
+        const errorObj = Error(JSON.stringify(error));
```
conswang, line 111: This syntax is probably correct, but just to be safe you might want to console log `errorObj` to make sure it matches up with what we're expecting? Apologies if you've already checked

frontend/hooks/data-fetching/course.ts
```diff
@@ -311,12 +310,13 @@ export async function createQuestion(
...
-        const error = await res.json();
-        throw {
-            payload: JSON.stringify(payload),
-            error: JSON.stringify(error),
-        };
+        if (res.status !== 409 && res.status !== 429) {
+            const error = await res.json();
+            logException(Error(JSON.stringify(error)), JSON.stringify(payload));
+        }
+        return new Promise((resolve) => resolve(res.status));
```
conswang: No need to manually construct a promise. You can just return res.status or undefined. Because the function is marked async, [the value will be wrapped in a promise anyway](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#:~:text=Async%20functions%20always%20return%20a%20promise.%20If%20the%20return%20value%20of%20an%20async%20function%20is%20not%20explicitly%20a%20promise%2C%20it%20will%20be%20implicitly%20wrapped%20in%20a%20promise.)

conswang (commented):
Everything looks good except for that one bug with await and .then both being used, the rest are just small fixes to consider. Thanks for fixing not one, but two bugs with our error response handling!

frontend/hooks/data-fetching/course.ts
```diff
@@ -330,10 +330,9 @@ export async function createQueue(
 
     if (!res.ok) {
         const error = await res.json();
-        throw {
-            payload: JSON.stringify(payload),
-            error: JSON.stringify(error),
-        };
+        const errorObj = Error(JSON.stringify(error));
+        logException(errorObj, JSON.stringify(payload));
```
eyingxuan, line 334: Not sure if this is planned in the future, but I'm not sure if it's correct to have inconsistent behavior between `createQueue` and `createQuestion`. In particular, `createQuestion` will no longer throw an exception while `createQueue` and `finishQuestion` will, and the return types are also different. Would it make sense for all these functions to return `Promise<[number, any]>`, where the number is the status code and any is the error object or undefined if it is a success? Could be even better if we define some type that is a discriminated union where a success status code does not require an error object. 

This does sound like a bigger refactor, and I don't mind letting this through for now.

eyingxuan (approved):
lgtm! Thanks for cleaning up my mess @danielxue

frontend/hooks/data-fetching/course.ts
danielxue: Theoretically we should do this but we decided to just refactor the code that is throwing errors for now, so I think this is getting put in the backlog for now.

frontend/components/Course/InstructorQueuePage/CreateQueue/CreateQueue.tsx
danielxue: Currently createQueue is implemented to throw an error for unexpected responses, and is currently getting caught to update an error state variable. I can change it to instead return a status promise similar to what I did with createQuestion to eliminate the try catch In general this is a case of refactoring the return to be consistently the status so that we don't have to catch the error.

conswang: Ah you're right, my bad, mixed up createQueue with createQuestion. Let's leave it like this

conswang (approved):
Thanks for addressing the changes! Let's merge it :>

PR #178, opened 2021-11-20
Add missing template rejection reason

ArmaanT:
This PR adds a question rejection reason to let students know that their question did not use the question template (correctly). Previously in 120, TAs were using NOT_SPECFIC which introduced some ambiguity over what wasn't specific. 

I'm not 100% sure I like this wording, but couldn't think of a better short phrase.


conswang:
How about "Did Not Use Template" or "Didn't Use Template"? "Didn't Follow Template"?

conswang (approved):

ArmaanT:
@KeenWill any preferences on wording?

KeenWill:
I think "Didn't Follow Template" is my favorite

PR #180, opened 2021-12-11
Feature/calendar-backend

bqle:
Calendar feature for students & TAs to schedule & update events happening in the course


backend/ohq/permissions.py
```diff
@@ -392,3 +395,66 @@ def has_permission(self, request, view):
...
+
+    def has_permission(self, request, view):
+        # Anonymous users can't do anything
+        if not request.user.is_authenticated:
+            return False
+
+        if view.action in ["create", "partial_update", "update"]:
+            course_pk = json.loads(request.body)["courseId"]
+            course = Course.objects.get(pk=course_pk)
+            membership = Membership.objects.filter(course=course, user=request.user).first()
```
ArmaanT: This actually isn't a safe check to make. It assumes that the DRF renderer is the custom camel case one, but if that ever changes this code will throw a KeyError. Instead you can replace this with the following line (which we already use for other permission classes)

```
membership = Membership.objects.get(course=view.kwargs["course_pk"], user=request.user)
````

Comment (bqle) [reply backend/ohq/permissions.py]: Hi Armaan! Thanks for the speedy reply. Since this courseId is posted in the json body, I don't know how to access it using view.kwargs. I tried what you suggested but I think view.kwargs only access the pk if it is in the url. Do you know what the best way is to grab info from a POST body?

**Comment by bqle** [comment]
Hi guys, the codecov I'm failing are only from the Schemas I created in views.py. They are basically for documentation and not actual logic code. Do we have ways of bypassing this?

File: backend/ohq/permissions.py [hunk 2]
```diff
@@ -392,3 +395,86 @@ def has_permission(self, request, view):
...
+
+        if view.action in ["create"]:
+            course_pk = json.loads(request.body)["courseId"]
+            course = Course.objects.get(pk=course_pk)
+            membership = Membership.objects.filter(course=course, user=request.user).first()
+            if membership is None:
+                return False
+            return membership.is_ta
+
+        if view.action in ["partial_update", "update"]:
```
kevc528: Move these update permissions and the retrieve and destroy permissions to has_object_permissions

backend/ohq/permissions.py
```diff
@@ -392,3 +395,86 @@ def has_permission(self, request, view):
...
+
+class OccurrencePermission(permissions.BasePermission):
+    def get_membership_from_event(self, request, event):
+        event_course_relation = EventRelation.objects.filter(event=event).first()
+        membership = Membership.objects.filter(
+            course_id=event_course_relation.object_id, user=request.user
+        ).first()
+        return membership
+
+    def has_object_permission(self, request, view, obj):
```
kevc528, line 456: Might want to associate event with a TA, and only allow the specific TA who created the event, head TA's, and prof. to edit/delete the event

backend/ohq/permissions.py
kevc528: Also, might want to associate event with a TA, and only allow the specific TA who created the event, head TA's, and prof. to edit/delete the event

backend/ohq/permissions.py
```diff
@@ -392,3 +395,91 @@ def has_permission(self, request, view):
...
+
+        if view.action in ["list"]:
+            course_ids = request.GET.getlist("course")
+            for course in course_ids:
+                membership = Membership.objects.filter(course=course, user=request.user).first()
+                if membership is None:
+                    return False
+            return True
+
+        if view.action in ["partial_update", "update", "retrieve", "destroy"]:
```
kevc528: If you look at this (https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions), we know that the has object permissions function will be called after the has permission function will is called, so we don't have to explicitly call it here. Just make sure that we return True at the end of this function.

backend/ohq/permissions.py
```diff
@@ -392,3 +395,91 @@ def has_permission(self, request, view):
...
+            # if any member of the course in the list is not accessible, return false
+            course_ids = request.GET.getlist("course")
+            for course in course_ids:
+                membership = Membership.objects.filter(course=course, user=request.user).first()
+                if membership is None:
+                    return False
+            return True
+
+        if view.action in ["retrieve", "update", "partial_update"]: 
+           return self.has_object_permission(request, view, None)
```
kevc528: Same here thing as the previous comment on EventPermission

backend/ohq/permissions.py
```diff
@@ -392,3 +395,91 @@ def has_permission(self, request, view):
...
+        if view.action in ["retrieve"]:
+            occurrence = Occurrence.objects.filter(pk=view.kwargs["pk"]).first()
+            membership = self.get_membership_from_event(request=request, event=occurrence.event)
+            return membership is not None
+
+        if view.action in ["update", "partial_update"]:
+            occurrence = Occurrence.objects.filter(pk=view.kwargs["pk"]).first()
+            membership = self.get_membership_from_event(request, occurrence.event)
+            return membership is not None and membership.is_ta
+
```
kevc528, line 471: Consider adding a destroy permission as well

backend/ohq/serializers.py
```diff
@@ -447,3 +449,105 @@ def create(self, validated_data):
...
+            ("start" in validated_data and instance.start != validated_data["start"])
+            or ("end" in validated_data and instance.end != validated_data["end"])
+            or (
+                "end_recurring_period" in validated_data
+                and instance.end_recurring_period != validated_data["endRecurringPeriod"]
+            )
+            or ("rule" in validated_data and rule.frequency != validated_data["rule"]["frequency"])
+        ):
+            if "rule" in validated_data:
+                rule = Rule.objects.create(frequency=validated_data["rule"]["frequency"])
```
kevc528: For rules, consider using get_or_create so we don't create many duplicate rules

backend/ohq/serializers.py
```diff
@@ -447,3 +449,105 @@ def create(self, validated_data):
...
+
+        instance.rule = rule
+        instance.save()
+        return instance
+
+    def create(self, validated_data):
+        course = Course.objects.get(pk=validated_data["course_id"])
+        rule = None
+        if "rule" in validated_data:
+            rule = Rule.objects.create(frequency=validated_data["rule"]["frequency"])
```
kevc528: same here with get_or_create

backend/ohq/serializers.py
```diff
@@ -447,3 +449,105 @@ def create(self, validated_data):
...
+            if "rule" in validated_data:
+                rule = Rule.objects.create(frequency=validated_data["rule"]["frequency"])
+                validated_data.pop("rule")
+            Occurrence.objects.filter(event=instance).delete()
+
+        # can never change course_id, client should create a new event instead
+        validated_data.pop("course_id")
+        super().update(instance, validated_data)
+
+        instance.rule = rule
```
kevc528, line 531: are we able to include rule in validated_data?

backend/ohq/serializers.py
```diff
@@ -447,3 +449,105 @@ def create(self, validated_data):
...
+class EventSerializer(serializers.ModelSerializer):
+    """
+    Serializer for events
+
+    All times are converted to UTC+0
+    """
+
+    rule = RuleSerializer(required=False)
+    course_id = serializers.IntegerField(required=False)
+
```
kevc528, line 487: Just read this stack overflow post: https://stackoverflow.com/questions/53735960/django-rest-framework-how-to-make-field-required-read-only-only-for-update-ac/53736259

Seems like if we use this, we can make certain fields required for POST like the course id

backend/ohq/views.py
```diff
@@ -607,3 +614,162 @@ class AnnouncementViewSet(viewsets.ModelViewSet):
...
+            tzinfo=utc
+        )
+        courses = Course.objects.filter(pk__in=course_ids)
+        erm = EventRelationManager()
+        occurrences = []
+        for course in courses:
+            events_for_course = erm.get_events_for_object(course)
+            for event in events_for_course:
+                for occurrence in event.get_occurrences(filter_start, filter_end):
+                    occurrence.save()
```
kevc528, line 800: can you add a quick comment on why occurrence.save() needs to be called, for future reference

kevc528 (changes requested):

kevc528 (approved):

PR #182, opened 2022-01-18
Implement frontend for queue pin feature

danielxue:
Instructor side (not editing and editing): 
Screenshot 2022-01-18 at 16-28-35 OHQ ABC 101

Screenshot 2022-01-18 at 16-28-43 OHQ ABC 101

Student side: 
Screenshot 2022-01-18 at 16-28-50 OHQ ABC 101


frontend/components/Course/InstructorQueuePage/CreateQueue/CreateQueue.tsx
```diff
@@ -267,7 +280,25 @@ const CreateQueue = (props: CreateQueueProps) => {
                                 />
                             </Form.Group>
                         </Form.Field>
-
+                        <Form.Field required>
+                            <label htmlFor="pin-radio">PIN</label>
+                            <Form.Group id="pin-radio">
+                                <Form.Radio
+                                    label="Enable PIN"
```
conswang: Nit - I'd prefer to be consistent with spelling "Pin", not "PIN"

frontend/components/Course/InstructorQueuePage/CreateQueue/CreateQueue.tsx
```diff
@@ -159,6 +173,7 @@ const CreateQueue = (props: CreateQueueProps) => {
                                 onChange={handleInputChange}
                             />
                         </Form.Field>
+                        videoChatSetting
```
conswang: This added line is getting rendered as "videoChatSetting", looks like a typo
image

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -1,11 +1,15 @@
-import { useState, useEffect, useMemo } from "react";
+import React, { useState, useEffect, useMemo } from "react";
```
conswang: nit - don't need to import React

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -70,11 +108,72 @@ const Queue = (props: QueueProps) => {
...
     };
 
     const onClose = async () => {
+        setEditingPin(false);
         await mutate(queueId, { active: false });
+        await partialUpdateQueue(courseId, queueId, { pin: "" });
+        await mutate(pin, { pin: pinState });
+    };
+
+    const renderPinInput = () => {
```
conswang: This can just be a variable and not a function, eg. const renderPinInput = ....

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -87,17 +186,25 @@ const Queue = (props: QueueProps) => {
...
+                    <Header.Subheader
+                        style={{
+                            whiteSpace: "break-spaces",
+                            wordBreak: "break-word",
+                        }}
+                    >
+                        {queue.description}
+                    </Header.Subheader>
+                </Header>
+                {queue.pinEnabled && renderPinInput()}
```
conswang: Need extra check if queue.isOpen (sry might have a different name)
When queue is closed, it doesn't make sense to render the pin, it displays as an empty box
image

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -70,11 +108,72 @@ const Queue = (props: QueueProps) => {
...
+        await partialUpdateQueue(courseId, queueId, { pin: "" });
+        await mutate(pin, { pin: pinState });
+    };
+
+    const renderPinInput = () => {
+        const editAllowed = editingPin && active;
+        const textColor = editAllowed ? "black" : "#B9B9BA";
+        const val = editingPin || !pinMutated ? pinState : pin;
+        return (
+            <div style={{}}>
```
conswang: don't need style prop

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -70,11 +108,72 @@ const Queue = (props: QueueProps) => {
...
+                    }}
+                >
+                    Pin:
+                </label>
+                <input
+                    name="changePin"
+                    value={active ? val : ""}
+                    onChange={handlePinChange}
+                    onKeyDown={handlePinKeyPress}
+                    style={{
```
conswang: This css definitely needs to be simplified, sorry I don't have any quick fixes.

I'll try to rebuild the box you did (it looks great!) but try to boil down the styling to only what we need. Will send you the code later

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -87,17 +186,25 @@ const Queue = (props: QueueProps) => {
...
-                    {queue.description}
-                </Header.Subheader>
-            </Header>
+            <div
+                style={{
+                    display: "flex",
+                    justifyContent: "space-between",
+                }}
+            >
+                <Header as="h3">
```
conswang: Looks like something funky is going on with the spacing which wasn't before
image

I think -
before we had the title inside the grid, which provided spacing
now we're putting the title in a flexbox outside the grid, and need to add a margin? 

Not sure why it's not happening here though
image

Anyway just make it look good on chrome and we should be fine

frontend/components/Course/InstructorQueuePage/QueueSettings/QueueForm.tsx
```diff
@@ -342,6 +357,25 @@ const QueueForm = (props: QueueFormProps) => {
...
+                            <Form.Group id="pin-radio">
+                                <Form.Radio
+                                    label="Enable PIN"
+                                    checked={input.pinEnabled}
+                                    name="pinEnabled"
+                                    disabled={loading}
+                                    onChange={handlePinInputChange}
+                                />
+                                <Form.Radio
+                                    label="Disable PIN"
```
conswang: Nit - same thing, be consistent with "Pin" instead of "PIN"

frontend/constants.ts
```diff
@@ -1,10 +1,11 @@
 export const ANNOUNCEMENTS_POLL_INTERVAL = 10000;
-export const QUEUE_STATUS_POLL_INTERVAL = 30000;
+export const QUEUE_STATUS_POLL_INTERVAL = 10000;
```
conswang: just checking, we meant to change on purpose so we can have the queue pin UI be shown immediately when the queue opens right? With the longer poll interval, you were experiencing some delay?

frontend/hooks/data-fetching/course.ts
```diff
@@ -336,6 +336,28 @@ export async function createQueue(
...
+        `/api/courses/${courseId}/queues/${queueId}/`,
+        {
+            method: "PATCH",
+            body: payload,
+        }
+    );
+
+    if (!res.ok) {
+        const error = await res.json();
+        throw {
```
conswang: Not sure if we should throw an error in this function, maybe it would be better to just return false or something, but we might as well be consistent with the rest of the code

However if !res.ok we should also log the error to sentry with logException(..) (see other functions)

Also we should be throwing an error object created with Javascript's Error() constructor, not sure what will happen if you pass it this data shape instead

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -57,6 +64,37 @@ const Queue = (props: QueueProps) => {
...
+            setEditingPin(false);
+        }
+    };
+
+    const handlePinToggle = () => {
+        if (editingPin) {
+            setPinMutated(false);
+            setEditingPin(false);
+            partialUpdateQueue(courseId, queueId, { pin: pinState });
+            mutate(pin, { pin: pinState }).then(() => setPinMutated(true));
```
conswang: Important - From my testing, the changed pin does not display correctly after you edit it

What are we using setPinMutated for?

conswang (changes requested):
Thanks for handling this complicated feature!

There are some easy bugs to address (typos), and some hard (the queue pin doesn't update correctly after you change the pin - need to get this working before we merge)

Sorry I couldn't debug those issues in more depth but hopefully this can get you started!

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -77,28 +118,88 @@ const Queue = (props: QueueProps) => {
         await mutate(queueId, { active: false });
     };
 
+    const pinInput = (
+        <div className="ui input">
+            <label
+                htmlFor="changePin"
+                style={{
+                    fontWeight: "bold",
+                    padding: "0.5rem 0.5rem 0.5rem 2rem",
```
conswang: not a big deal at all, just fyi there's shorthand
https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties#:~:text=The%202%2Dvalue%20syntax%3A%20border%2Dwidth%3A%201em%202em%20%E2%80%94%20The%20first%20value%20represents%20the%20vertical%2C%20that%20is%20top%20and%20bottom%2C%20edges%2C%20the%20second%20the%20horizontal%20ones%2C%20that%20is%20the%20left%20and%20right%20ones.

frontend/components/Course/InstructorQueuePage/Queue.tsx
```diff
@@ -77,28 +118,88 @@ const Queue = (props: QueueProps) => {
...
+                style={{
+                    height: "2.7rem",
+                    color: editingPin ? "black" : "#B9B9BA",
+                    caretColor: editingPin ? "#75767F" : "transparent",
+                    borderColor: "#22242626",
+                    backgroundColor: editingPin ? "white" : "#F1F1F2",
+                    paddingRight: "2rem",
+                }}
+            />
+            <Button
```
conswang: There seems to be a bug when you click the checkmark after you're done editing, it sets editing pin to true again
It only happens if you click right on the icon

conswang (changes requested):
Thanks for making all the requested changes!

Just another bug to fix before we merge, clicking checkmark doesn't work. I'll see what's going on there

Would also appreciate if you could put the queue pin in its own component - the queue file is getting large and there's a useeffect that we wouldn't want to be run in say, a queue that doesn't have pin enabled

danielxue:
Update mobile responsiveness. 

Desktop: 
Screenshot 2022-01-30 at 15-38-14 OHQ ABC 101

Desktop split screen: 
Screenshot 2022-01-30 at 15-38-23 OHQ ABC 101

Mobile: 
Screenshot 2022-01-30 at 15-38-30 OHQ ABC 101

conswang (approved):
Nice work!! Merge it as soon as you're ready

PR #188, opened 2022-01-31
automatic hyperlink using react-linkify

bqle:
Added Linkify for announcements & questions for both instructors & students

### DIFF
--- a/frontend/components/Course/Announcements.tsx
+++ b/frontend/components/Course/Announcements.tsx
@@ -8,6 +8,7 @@ import {
     Icon,
     Button,
 } from "semantic-ui-react";
+import Linkify from "react-linkify";
 import { mutateResourceListFunction } from "@pennlabs/rest-hooks/dist/types";
 import { useMediaQuery } from "@material-ui/core";
 import { Announcement, BaseUser, NotificationProps } from "../../types";
@@ -233,7 +234,7 @@ const AnnouncementMessage = ({
                             marginBottom: "0px",
                         }}
                     >
-                        {announcement.content}
+                        <Linkify>{announcement.content}</Linkify>
                     </p>
                     <br />
                     <p style={{ color: "#666666", marginTop: "0px" }}>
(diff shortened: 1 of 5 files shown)


danielxue:
Clicking on link in announcement navigates directly to link while clicking on link in question opens link in new tab. Is this intended behavior?

bqle:
Oh yeah, that's something I wanted to ask you guys. Do you think it would be better to open on a new tab? (I think so).

frontend/components/Course/Announcements.tsx
```diff
@@ -233,7 +234,23 @@ const AnnouncementMessage = ({
                             marginBottom: "0px",
                         }}
                     >
-                        {announcement.content}
+                        <Linkify
```
conswang: The linkify boilerplate code is shared by a lot of components, I'd move it over to a new component that only takes in a string like <LinkedText string={announcement.content}/> since we probably want all our links to behave the same way anyway. You can put it in the components/common/ui folder

That way if you're making a decision like whether to open links in a new tab, you only have to change 1 file

frontend/components/Course/Announcements.tsx
```diff
@@ -233,7 +234,23 @@ const AnnouncementMessage = ({
...
+                            componentDecorator={(
+                                decoratedHref: string,
+                                decoratedText: string,
+                                key: number
+                            ) => (
+                                <a
+                                    href={decoratedHref}
+                                    key={key}
+                                    target="_blank"
```
conswang: If you want to add target="_blank" to open in new tab you may have to add rel="noopener noreferrer" for security reasons
https://perishablepress.com/fix-blank-target-vulnerability/ although modern browsers should automatically set this
https://mathiasbynens.github.io/rel-noopener/

Don't think it's a big deal if you don't put it (although it won't hurt), perhaps you can use this to catch TAs who are using internet explorer...

conswang (changes requested):
First front end ticket 😮 and awesome job solving the problem on your own!

Just some small changes to make + it'd be nice if you could add a test plan

Uh unfortunately you can't escape the testing even in front end haha but instead of writing test cases you can just upload screenshots or screen recording if necessary for you and us to keep track of what's working

PR #190, opened 2022-02-04
Add character counts for course title

danielxue:
Implement character count for course title

Create: 
Screenshot 2022-02-04 at 18-05-36 OHQ

Edit: 
Screenshot 2022-02-04 at 18-04-56 OHQ CIS 199


frontend/components/Home/Dashboard/Forms/CreateCourseForm.tsx
```diff
@@ -53,9 +68,21 @@ const CreateCourseForm = (props: CreateCourseFormProps) => {
...
                     name="courseTitle"
-                    onChange={changeFunc}
+                    value={input.courseTitle}
+                    onChange={handleCourseTitleChange}
                     placeholder="Data Structures and Algorithms"
                 />
+                <div
+                    style={{
+                        textAlign: "right",
+                        color:
```
conswang, line 78: This is default color (black) when you don't hit the char limit right?

Not sure if you already knew but, you can also do fancy syntax like
style={{
textAlign: "right",
...courseTitleCharCount < COURSE_TITLE_CHAR_LIMIT && { color: crimson }
}}

conswang (approved):
Looks good, thanks for the fix!

Only thing that bothers me is that this design is only used in this one input field, the others look different. But I'm thinking that's a bigger issue and we really need to refactor all our forms to look consistent... Evidence is that you had to change this component in both create/edit course :<

Another idea is you could make the x/y char limit label a UI component and use this in the other fields

PR #199, opened 2022-03-20
Summary card analytics

danielxue:
Full page: 
Screenshot 2022-03-27 at 13-43-01 OHQ LABS 621

Tests: 
No entries:
Screenshot 2022-03-27 at 13-22-46 OHQ LABS 621

Mismatched date on students helped and average time per student: 
[
    {
        "metric": "STUDENTS_HELPED",
        "day": null,
        "hour": null,
        "value": "4.00000000",
        "date": "2022-03-27"
    }
]
[
    {
        "metric": "AVG_TIME_HELPING",
        "day": null,
        "hour": null,
        "value": "120.00000000",
        "date": "2022-03-26"
    }
]
Screenshot 2022-03-27 at 13-27-58 OHQ LABS 621
 
Average 1 entry: 
[
    {
        "metric": "STUDENTS_HELPED",
        "day": null,
        "hour": null,
        "value": "2.00000000",
        "date": "2022-03-26"
    }
]
[
    {
        "metric": "AVG_TIME_HELPING",
        "day": null,
        "hour": null,
        "value": "120.00000000",
        "date": "2022-03-26"
    }
]

Screenshot 2022-03-27 at 13-31-27 OHQ LABS 621

Average 2 entries: 
[
    {
        "metric": "STUDENTS_HELPED",
        "day": null,
        "hour": null,
        "value": "2.00000000",
        "date": "2022-03-26"
    },
    {
        "metric": "STUDENTS_HELPED",
        "day": null,
        "hour": null,
        "value": "4.00000000",
        "date": "2022-03-27"
    }
]
[
    {
        "metric": "AVG_TIME_HELPING",
        "day": null,
        "hour": null,
        "value": "120.00000000",
        "date": "2022-03-26"
    },
    {
        "metric": "AVG_TIME_HELPING",
        "day": null,
        "hour": null,
        "value": "480.00000000",
        "date": "2022-03-27"
    }
]

Screenshot 2022-03-27 at 13-34-18 OHQ LABS 621

Outside date range: 
[
    {
        "metric": "STUDENTS_HELPED",
        "day": null,
        "hour": null,
        "value": "2.00000000",
        "date": "2022-03-26"
    },
    {
        "metric": "STUDENTS_HELPED",
        "day": null,
        "hour": null,
        "value": "4.00000000",
        "date": "2022-03-19"
    }
]
[
    {
        "metric": "AVG_TIME_HELPING",
        "day": null,
        "hour": null,
        "value": "120.00000000",
        "date": "2022-03-26"
    },
    {
        "metric": "AVG_TIME_HELPING",
        "day": null,
        "hour": null,
        "value": "480.00000000",
        "date": "2022-03-21"
    }
]

Screenshot 2022-03-27 at 13-42-20 OHQ LABS 621
Screenshot 2022-03-27 at 13-42-34 OHQ LABS 621


frontend/components/Course/Analytics/Analytics.tsx
```diff
@@ -177,45 +178,44 @@ const Analytics = ({ course, queues }: AnalyticsProps) => {
...
+                            inline
+                            options={queueOptions}
+                            defaultValue={queueId}
+                            onChange={(e, { value }) => {
+                                setQueueId(value as number);
+                            }}
+                        />
+                        <SummaryCards courseId={course.id} queueId={queueId} />
+                        <Averages courseId={course.id} queueId={queueId} />
+                        {/* <Segment basic>
```
conswang: Can delete this and other commented code, the <MyPieChart/> component is already gone

frontend/components/Course/Analytics/Analytics.tsx
```diff
@@ -177,45 +178,44 @@ const Analytics = ({ course, queues }: AnalyticsProps) => {
...
-                            <Dropdown
-                                inline
-                                options={queueOptions}
-                                defaultValue={queueId}
-                                onChange={(e, { value }) => {
-                                    setQueueId(value as number);
-                                }}
-                            />
-                        </span>
+                        You have no queues. Create a queue on the{" "}
```
conswang, line 45: Test that (no queues) this still looks okay

frontend/hooks/data-fetching/analytics.ts
```diff
@@ -72,3 +78,32 @@ export const useHeatmapData = (
...
+    type: Metric,
+    range: number
+) => {
+    let dateRange = "";
+    if (range) {
+        let date = new Date();
+        date.setDate(date.getDate() - range);
+        date = new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000);
+        const formattedDate = date.toISOString().split("T")[0];
+        dateRange = range ? `&date__gte=${formattedDate}` : "";
```
conswang: You already checked if range === 0 in line 89 so no need to check again

However I'm a bit confused on what why range could be 0 (the only falsy number value)? Looks like we're only using 7, 30, or 90

Does 0 mean 1 day of stats? If so would probably make range an enum

frontend/hooks/data-fetching/analytics.ts
```diff
@@ -72,3 +78,32 @@ export const useHeatmapData = (
...
+    }
+    const { data, error, isValidating } = useResource<AnalyticsData>(
+        `/api/courses/${courseId}/queues/${queueId}/statistics/?metric=${type}${dateRange}`,
+        {
+            revalidateOnFocus: false,
+            revalidateOnReconnect: false,
```
conswang, line 102: Nice, you can also put revalidateIfStale: false

Found this from https://swr.vercel.app/docs/revalidation

There seems to be a useSWRImmutable hook but that's not available in frontend-core, so let's just use these 3 options. Also, make these options a shared variable in the analytics file like `immutableOptions` or something and add them to the heatmap useResource call, thanks :>

frontend/hooks/data-fetching/analytics.ts
```diff
@@ -72,3 +78,32 @@ export const useHeatmapData = (
...
+        }
+    );
+
+    if (data) {
+        return { data, error, isValidating };
+    }
+
+    return { data: [], error, isValidating };
```
conswang: save a few lines of code if you want
return { data: data || [], error, isValidating };

frontend/components/Course/Analytics/Cards/SummaryCards.tsx
```diff
@@ -0,0 +1,128 @@
...
+                        setTimeRange(value as number);
+                    }}
+                />
+            </div>
+            <Card.Group>
+                <AnalyticsCard
+                    label="Questions Answered"
+                    content={
+                        numAnsweredValidating
+                            ? "..."
```
conswang: Might be just a little cleaner to separate out all the display logic into AnalyticsCard, aka. pass isValidating as a prop to the card component and conditionally display "..." there

frontend/components/Course/Analytics/Cards/SummaryCards.tsx
```diff
@@ -0,0 +1,128 @@
...
+        isValidating: avgTimeHelpingValidating,
+    } = useStatistic(courseId, queueId, Metric.AVG_TIME_HELPING, timeRange);
+
+    const sum = (data) => {
+        return data.reduce((prev, cur) => prev + parseInt(cur.value, 10), 0);
+    };
+
+    const averageWeighted = (data, weights) => {
+        const timeHelping: { value: number; date: string }[] = [];
+        const numHelped: { value: number; date: string }[] = [];
```
conswang: ~~Honestly not much of an efficiency boost with only 90 items but it might be more readable to~~
~~- put data into a map, with key = dataItem.date and value = dataItem.value~~
~~- put weights into a map also with key = weightItem.date and value = weightItem.value~~
~~- loop through entryset and calculate sums~~

~~Actually... you don't have to rewrite the whole thing unless your CIS 121 OCD is bothering you...~~

Sorry ignore the above, overhead definitely not worth it

You can just get rid of the timeHelping and numHelped arrays and keep instead a dataSum += dataItem.value * weightItem.value and  weightSum += weightItem.value right?

Haha hopefully I didn't horribly misread the code

frontend/components/Course/Analytics/Cards/SummaryCards.tsx
```diff
@@ -0,0 +1,128 @@
...
+                    value: dataItem.value * weightItem.value,
+                    date: dataItem.date,
+                });
+                numHelped.push({
+                    value: weightItem.value,
+                    date: dataItem.date,
+                });
+            }
+        }
+        return sum(timeHelping) / sum(numHelped) || 0;
```
conswang: Why || 0? If you're checking for divide by 0 errors (which you should definitely do!) I think javascript returns infinity which is not falsy :(

conswang (changes requested):
Appreciate the hard work! As usual, implementing a "simple" feature is more effort than expected ahaha

Most of the comments are small nitpicks which you can address as you fit, only 2 things that we want more clarification/fixing on before merging
- what the range parameter could be in the useStatistic function, aka. will range = 0 have some special meaning?
- clean up the averageWeighted function in the SummaryCards component

frontend/components/Course/Analytics/Cards/SummaryCards.tsx
danielxue: interesting, so it looks like js treats 0/0 as NaN which is falsy, but any other numerator as infinity. i will still change it to an explicit check though in the edge case some happens on the backend such that we have a mismatch and timeHelping is positive when numHelped is zero.

frontend/hooks/data-fetching/analytics.ts
danielxue: Yep I found it from the same link. For some reason though, it doesn't work. I think revalidateIfStale isn't in the typescript definition (possibly a bug with next). Getting TS2345: Argument of type '{ revalidateOnFocus: false; revalidateOnReconnect: false; revalidateIfStale: boolean; }' is not assignable to parameter of type 'ConfigInterface<AnalyticsData, any, fetcherFn<AnalyticsData>>'.

frontend/hooks/data-fetching/analytics.ts
danielxue: Good catch. I initially implemented range as an optional parameter, allowing undefined (no range would give all entries, rather than all those >= a date). Not sure why I changed that, but do you think that I should add that back, or perhaps reserve 0 as that, or reimplement as a enum?

frontend/components/Course/Analytics/Cards/SummaryCards.tsx
danielxue: yep you're right. definitely overcomplicated this.

frontend/hooks/data-fetching/analytics.ts
conswang: Oh this seems to be missing from our front end core package, just ignore it, will make a note to update frontend core

frontend/hooks/data-fetching/analytics.ts
conswang: Range can only be 7, 30, and 90 then? In that case you could use an enum. No need to reserve special range values like 0 or undefined since we're not using them in any components yet, so you can remove the range = 0 checks

frontend/components/Course/Analytics/Cards/SummaryCards.tsx
conswang: Sounds good! Also, if you get an error, just log it with sentry and maybe don't display the card - you can return undefined in this helper function and check if average is undefined or not before you display the card. Or display some error state! Not a big deal what you choose, but definitely at least log the error instead of silently returning 0 as the average wait time

frontend/hooks/data-fetching/analytics.ts
danielxue: Should I not make the hook more general though? in case we do make a component that requires a different range from that. at the very least i coiuld not deal with 0 or undefined and just require a positive range to simplify things but I'm not sure what the benefit of using an enum would be.

conswang (approved):
Nice!! Let's get this one in on April 2nd haha

PR #206, opened 2022-04-01
Modify DOMAIN handling

ArmaanT:
This PR modifies the backend to load accepted domains from an environment variable called `DOMAINS` that is comma delimited.


joyliu-q (approved):
LGTM! 🥳

PR #214, opened 2022-05-26
Feature/file upload

bqle:
Students can now upload files and delete files associated with questions.


backend/ohq/views.py
```diff
@@ -342,6 +350,57 @@ def quota_count(self, request, course_pk, queue_pk):
         else:
             return JsonResponse({"detail": "queue does not have rate limit"}, status=405)
 
+    def handle_file_upload(self, uploaded_file, question):
+        question_file = QuestionFile(question=question, file=uploaded_file)
+        question_file.save()
+        return question_file
```
bqle, line 356: Will this automatically save to S3 instead of local?

PR #224, opened 2022-10-07
Use Shared Actions

joyliu-q:
Use shared CI/CD from Penn Labs' [shared actions](https://github.com/pennlabs/shared-actions).

With this, developers can gain experience with using GitHub workflows, while maintaining the abstractions that Kraken provided.


joyliu-q (commented):
LGTM!

PR #233, opened 2022-10-15
allow head ta finish question

vcai122:
is this good


vcai122:
fixes #233

bqle:
Hi Vincent! This will allow Head TAs to delete the question from our database. The way OHQ handles questions answered is by PATCHING the status field of the Question model to one of the STATUS_CHOICES. I actually made a mistake on the issue because it seems this function is already supported. Thank you for checking out our codebase; feel free to check out our other issues as well.

PR #240, opened 2022-12-04
Use shared actions

alxkp:
Updated version of reusable actions in workflow to v0.1 release of shared-actions


.github/workflows/shared-build-deploy.yaml
```diff
@@ -0,0 +1,88 @@
+
+# ========================================
```
joyliu-q, line 1: Thanks for adding this comment!

joyliu-q (approved):
LGTM!

Issue #255, opened 2023-02-11
[Dependencies] Bump node version to v18


danielxue:
Node v14 is reaching EOL (end of April). We should upgrade to v18 LTS. Only change should be frontend Dockerfile config.

PR #257, opened 2023-03-17
Upgrade node to version 18

danielxue:
Must be merged after bumping next to v12 (to avoid openSSL issues)


danielxue:
Closes #255

PR #266, opened 2023-05-02
arching spring 23

bqle:


.github/workflows/shared-build-deploy.yaml
```diff
@@ -7,14 +7,14 @@ name: Build and Deploy OHQ
...
 jobs:
-  backend-check:
-    name: "Backend Check"
-    uses: pennlabs/shared-actions/.github/workflows/django-check.yaml@v0.1
-    with:
-      projectName: officehoursqueue
-      path: backend
-      flake: true
-      black: true
+  # backend-check:
```
joyliu-q, line 10: Any reason we are choosing to not do the backend checks for OHQ from now on?

bqle: I just wanted to remove the codecov command that was failing [here](https://github.com/pennlabs/office-hours-queue/actions/runs/4877624280/jobs/8702437157), but didn't want to change the shared-actions.

Can you selectively disable codecov for us Joy?

PR #272, opened 2023-05-11
Version updates and fixes

danielxue:
React v17.0.2 -> v18.2.0
- Fixed hydration error with notification support check unwrapped in useEffect

Next v12.1.0 -> v13.4.2
- Fixed next Link issues due to migration


danielxue:
This commit has sort of ballooned. Most changes are very simple linting fixes that arose due to upgrading linting package versions (necessary since next upgrade => typescript upgrade => linter upgrades). Would be good to get a second set of eyes to make sure that I didn't inadvertently change behavior of anything/introduce bugs, and that none of the dep changes (next 13/react 18/typescript 5) break things.

PR #285, opened 2023-10-01
Update backend initial setup

trangiabach:
This PR:
- There were some bugs related to the postgres instance not starting or lacking an user when setting up the backend so included some guides for that 👌!


joyliu-q (approved):
LGTM! Thanks for adding this Bach :D

joyliu-q:
Approving this to unblock u so this can get merged.

PR #288, opened 2023-10-14
feat(banner):maintenance

esinx:
Screenshot 2023-10-14 at 5 47 00 PM

### DIFF
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -1,9 +1,11 @@
-import { useEffect } from "react";
 import "semantic-ui-css/semantic.min.css";
 import "../styles/index.css";
-import { SWRConfig } from "swr";
+
+import { useEffect } from "react";
 // eslint-disable-next-line import/no-extraneous-dependencies -- weird, but eslint doesn't like this
 import { Announcement } from "@esinx/pennlabs-ui";
+import { SWRConfig } from "swr";
+
 import { doApiRequest } from "../utils/fetch";
 import withGA from "../utils/ga/withGA";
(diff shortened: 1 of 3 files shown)


frontend/pages/_app.tsx
```diff
@@ -2,6 +2,8 @@ import { useEffect } from "react";
 import "semantic-ui-css/semantic.min.css";
 import "../styles/index.css";
 import { SWRConfig } from "swr";
+// eslint-disable-next-line import/no-extraneous-dependencies -- weird, but eslint doesn't like this
```
joyliu-q: Is this something moving this dep to another section could fix? As you can see, we're doing relative imports above this line and below, so can we try moving this to the top?

frontend/pages/_app.tsx
```diff
@@ -11,17 +13,30 @@ const MyApp = ({ Component, pageProps }) => {
...
+        <>
+            <SWRConfig
+                value={{
+                    fetcher: (path, ...args) =>
+                        doApiRequest(path, ...args).then((res) => res.json()),
+                    refreshWhenHidden: true,
+                }}
+            >
+                {/* This is necessary for generic HOC */}
+                {/* eslint-disable-next-line */}
```
joyliu-q: Can we do specific eslint disabling for the next line (for a specific warning) so there's more clarity?

joyliu-q (changes requested):
A few eslint warning disabling changes. Also, are the `doApiRequest` correctly configured, because we're hardcoding it for this current maintenance iteration?

PR #290, opened 2023-10-25
Frontend/tip card

kygchng:


backend/Pipfile
```diff
@@ -20,7 +20,7 @@ dj-database-url = "*"
 djangorestframework = "*"
 psycopg2 = "*"
 sentry-sdk = "*"
-django = "==3.1.7"
+django = "==3.2.4"
```
danielxue: In general, please don't edit backend dependencies in a frontend change unless it's relevant to the frontend change. 

You should revert this change and the lock file change

frontend/components/Home/Dashboard/tips.json
```diff
@@ -0,0 +1,27 @@
...
+        "title": "FAQ",
+        "description": "Have questions? Look through our “FAQ” tab on the sidebar!",
+        "student": true
+    },
+    {
+        "title": "[Instructor] View Past Questions",
+        "description": "Go to Question Summary to view past questions.",
+        "student": false
+    }
+]
```
danielxue: nit: add trailing newline at end of file

frontend/components/Home/Dashboard/Dashboard.tsx
```diff
@@ -17,18 +17,16 @@ import {
 } from "../../../constants";
 import ModalShowNewChanges from "./Modals/ModalShowNewChanges";
 import updatedMd from "../../Changelog/changelogfile.md";
+import tips from "./tips.json";
+import { transpileModule } from "typescript";
```
danielxue: I think this is accidentally imported? If it's not necessary please remove it.

frontend/components/Home/Dashboard/Dashboard.tsx
```diff
@@ -62,6 +60,40 @@ const Dashboard = () => {
     const [logOpen, setLogOpen] = useState(false);
     const [logModal, setLogModal] = useState(false);
 
+    const [messageDisp, setMessageDisp] = useState(false);
+   
+    const [tipCard, setTipCard] = useState({title: "", description: "", student: false});
```
danielxue: I was thinking about the design of this a bit more. Strictly, it's unnecessary for the tip to be stored as state. In general, if you can calculate a value on the fly,  you should do that rather than rely on state (this prevents inadvertent issues with inconsistent state or a blank tip card first loading and then populating with a tip). 

So the best approach is to define a function getTip() which would have the same logic as in your useEffect hook but then just return the tip message, and call that whenever you want to display the tip.

frontend/components/Home/Dashboard/Dashboard.tsx
```diff
@@ -62,6 +60,40 @@ const Dashboard = () => {
     const [logOpen, setLogOpen] = useState(false);
     const [logModal, setLogModal] = useState(false);
 
+    const [messageDisp, setMessageDisp] = useState(false);
```
danielxue: messageDisp logic can remain in the useEffect hook (see below comment). Although strictly calculating it on the fly with a useMemo hook might be better but it's whatever.

frontend/components/Home/Dashboard/Dashboard.tsx
```diff
@@ -62,6 +60,40 @@ const Dashboard = () => {
...
+            //instructor
+            const currentDate = new Date();
+            const index = (currentDate.getDate())  % tips.length;
+            setTipCard(tips[index]);
+        }
+
+        const currentDate = new Date();
+        const today = currentDate.toDateString();
+        
+        console.log(today);
```
danielxue: Shouldn't have console logs in production-ready code.

frontend/components/Home/Dashboard/Dashboard.tsx
```diff
@@ -62,6 +60,40 @@ const Dashboard = () => {
...
+        console.log(localStorage.getItem("lastSeenDate"));
+        if (!(today === localStorage.getItem("lastSeenDate"))) {
```
danielxue: This conditional can be simplified down to just setMessageDisp(today !== localStorage.getItem("lastSeenDate");

frontend/components/Home/Dashboard/Dashboard.tsx
```diff
@@ -62,6 +60,40 @@ const Dashboard = () => {
     const [logOpen, setLogOpen] = useState(false);
     const [logModal, setLogModal] = useState(false);
 
+    const [messageDisp, setMessageDisp] = useState(false);
+   
+    const [tipCard, setTipCard] = useState({title: "", description: "", student: false});
+
+    var tipsArr = tips;
+
+    useEffect(() => {
```
danielxue, line 67: Since there's a not insubstantial logic associated with the tip card, it might be best to spool it out into it's own component.

danielxue (changes requested):

danielxue (approved):

PR #293, opened 2023-10-29
Backend/llm responses

benjmnxu:
pr request


trangiabach:
> pr request

Encourage writing some tests to make sure that LLM responses and settings works correctly with each CRUD operations

Pipfile
trangiabach, line 1: I think you should remove this global Pipfile

frontend/yarn.lock
trangiabach, line 1: I think you should reset this yarn.lock file since we are not touching any files in the frontend

backend/ohq/views.py
```diff
@@ -774,3 +778,114 @@ def list(self, request, *args, **kwargs):
...
+            IF  YOU DO NOT FULLY UNDERSTAND THE USER'S QUESTION, ASK A FOLLOW UP QUESTION
+
+            YOU WOULD MUCH RATHER ANSWER TOO LITTLE THAN TOO MUCH.
+
+            Below are examples, where the question is deliminated in ticks, your response is in brackets, and each example is separated by commas:
+            'Q4?': [Please elaborate more on your question if you would like me to help]
+        """
+        return super().create(request, *args, **kwargs)
+    
+    def update(self, request, *args, **kwargs):
```
trangiabach: You should implement this update method and it should not replace but only append to the current LLM prompt

benjmnxu:
New update to pr

benjmnxu:
New Update to pr

PR #294, opened 2023-11-05
Paginate roster

kunlizhang:
This PR paginates the roster page to display just 20 people at the same time. In the case where a filter is applied, it resets to the 1st page and also still displays 20 at the same time.
Screenshot 2023-11-05 at 1 07 39 PM

### DIFF
--- a/frontend/components/Course/Roster/Roster.tsx
+++ b/frontend/components/Course/Roster/Roster.tsx
@@ -488,23 +488,14 @@ const Roster = (props: RosterProps) => {
                                                     event,
                                                     { activePage }
                                                 ) => {
-                                                    let parsedPage: number;
-                                                    if (
+                                                    const parsedPage: number =
                                                         typeof activePage ===
                                                         "string"
-                                                    ) {
-                                                        parsedPage = parseInt(
-                                                            activePage,
-                                                            10
-                                                        );
-                                                    } else if (
-                                                        typeof activePage ===
-                                                        "number"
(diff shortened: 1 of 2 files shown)


frontend/components/Course/Roster/Roster.tsx
```diff
@@ -464,6 +471,49 @@ const Roster = (props: RosterProps) => {
...
+                                            <Pagination
+                                                activePage={activePageState}
+                                                totalPages={Math.max(
+                                                    1,
+                                                    Math.ceil(
+                                                        filteredUsers.length /
+                                                            MAX_MEMBERS_PER_PAGE
+                                                    )
+                                                )}
+                                                onPageChange={(
```
danielxue, line 487: Can you clean up this function and also update the summary table with the same? There really should be a one or two line equivalent to this.

danielxue: At the very least if you set activePage's type to number | string | undefined you can just set it directly. Summary might be a bit weird though but take a look at that well.

danielxue (changes requested):

frontend/components/Course/Roster/Roster.tsx
kunlizhang: Refactored with fancy ternary function. Must define new variable as linter isn't happy if we redefine function parameter.

danielxue (approved):

PR #303, opened 2024-02-18
TA Reviews Improvements (Frontend + Backend)

trangiabach:
- Frontend + Backend Integration Improvements for TA reviews


backend/ohq/migrations/0001_initial.py
rm03, line 1: Can you explain why all migration files are being merged into one here?

trangiabach: Good catch! There were some conflict that was happening on my local computer that needed this change but yes I was intending to make them separated in later commits!

trangiabach: Forgot to mention this. This PR is still WIP, let me make it a draft!

PR #308, opened 2024-03-24
[Improvements] Migrate to Django 5

trangiabach:
This PR:
- Migrates `django` from 3.1.7 to 5.0.3
- Migrates `channels` from <=3 to 3.0.5 (higher version will break `django-rest-live` used for the web socket)
- Migrates Python to `3.11`. Change base image to support this version


judtinzhang (approved):

benjmnxu (commented):
could you just check why the pipfile.lock is conflicting. probably an issue with some new packages installed for calendar

PR #309, opened 2024-03-27
Add User Lifetime Statistics

rachllee:
Added a UserStatistics model & relevant methods to track a user's lifetime statistics across all classes 

Currently includes:
- lifetime number of questions asked
- lifetime number of questions answered
- lifetime amount of time spent receiving help
- lifetime amount of time spent helping
- lifetime number of students helped

and I tested them too :)


trangiabach:
> Added a UserStatistics model & relevant methods to track a user's lifetime statistics across all classes
> 
> Currently includes:
> 
> * lifetime number of questions asked
> * lifetime number of questions answered
> * lifetime amount of time spent receiving help
> * lifetime amount of time spent helping
> * lifetime number of students helped
> 
> and I tested them too :)

backend god @rachllee

backend/ohq/models.py
```diff
@@ -411,3 +411,35 @@ class Announcement(models.Model):
     author = models.ForeignKey(User, related_name="announcements", on_delete=models.CASCADE)
     time_updated = models.DateTimeField(auto_now=True)
     course = models.ForeignKey(Course, related_name="announcements", on_delete=models.CASCADE)
+
+
+class UserStatistic(models.Model):
```
judtinzhang, line 416: looks good

backend/package-lock.json
```diff
@@ -0,0 +1,27 @@
+{
```
judtinzhang: not sure what package-lock is here? should remove

backend/package.json
```diff
@@ -0,0 +1,5 @@
+{
```
judtinzhang: ditto

backend/ohq/statistics.py
```diff
@@ -232,3 +233,69 @@ def queue_calculate_questions_per_ta_heatmap(queue, weekday, hour):
...
+    num_questions = Decimal(Question.objects.filter(asked_by=user).count())
+    UserStatistic.objects.update_or_create(
+        user=user,
+        metric=UserStatistic.METRIC_TOTAL_QUESTIONS_ASKED,
+        defaults={"value": num_questions},
+    )
+
+
+def user_calculate_questions_answered(user):
+    num_questions = Decimal(Question.objects.filter(responded_to_by=user).count())
```
judtinzhang: nit: should add `status=Question.STATUS_ANSWERED` to make the query more accurate

backend/ohq/statistics.py
```diff
@@ -232,3 +233,69 @@ def queue_calculate_questions_per_ta_heatmap(queue, weekday, hour):
...
+        user=user,
+        metric=UserStatistic.METRIC_TOTAL_QUESTIONS_ANSWERED,
+        defaults={"value": num_questions},
+    )
+
+
+
+def user_calculate_time_helped(user):
+    user_time_helped = (
+        Question.objects.filter(asked_by=user)
```
judtinzhang: Does this query work if the question was rejected or withdrawn? In that case the `time_response_started` field would be null. If it doesn't, perhaps filter for where the question status is answered?

backend/ohq/statistics.py
```diff
@@ -232,3 +233,69 @@ def queue_calculate_questions_per_ta_heatmap(queue, weekday, hour):
...
+        .aggregate(time_helped=Sum(F("time_responded_to") - F("time_response_started")))
+    )
+    time = user_time_helped["time_helped"]
+
+    UserStatistic.objects.update_or_create(
+        user=user,
+        metric=UserStatistic.METRIC_TOTAL_TIME_BEING_HELPED,
+        defaults={"value": time.seconds if time else Decimal('0.00000000')},
```
judtinzhang: I am more inclined to not call `update_or_create` at all if `time.seconds` is `None` instead of updating the value to be 0.

backend/ohq/statistics.py
```diff
@@ -232,3 +233,69 @@ def queue_calculate_questions_per_ta_heatmap(queue, weekday, hour):
...
+    )
+
+
+def user_calculate_time_helping(user):
+    user_time_helping = (
```
judtinzhang: ditto for comments regarding `user_calculate_time_helped`

backend/ohq/management/commands/user_stat.py
```diff
@@ -0,0 +1,67 @@
...
+                course_questions = Question.objects.filter(queue__course=course)
+                date = (
+                    timezone.template_localtime(
+                        course_questions.earliest("time_asked").time_asked
+                    ).date()
+                    if course_questions
+                    else yesterday
+                )
+
+            course_questions = Question.objects.filter(queue__course=course, time_asked__gte=date)
```
judtinzhang:
```python3
from django.db.models import Q

questions_queryset = Question.objects.filter(queue__course=course, time_asked__gte=date)
users_union = (
    User.objects.filter(
        Q(id__in=questions_queryset.values_list("asked_by", flat=True)) |
        Q(id__in=questions_queryset.values_list("responded_to_by", flat=True))
    )
)
```

backend/ohq/management/commands/user_stat.py
```diff
@@ -0,0 +1,67 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument("--hist", action="store_true", help="Calculate all historic statistics")
+
+    def calculate_statistics(self, profiles, courses, earliest_date):
```
judtinzhang: The `earliest_date` parameter here is used to compute a `date` field, which I don't think is quite necessary. I think we can directly just use the `earliest_date` field wherever `date` is used.

backend/ohq/management/commands/user_stat.py
```diff
@@ -0,0 +1,67 @@
...
+                    if course_questions
+                    else yesterday
+                )
+
+            course_questions = Question.objects.filter(queue__course=course, time_asked__gte=date)
+            for q in course_questions:
+                active_users_today.add(q.asked_by)
+                active_users_today.add(q.responded_to_by)
+
+        for profile in profiles:
```
judtinzhang: This might be slow because we are iterating over all profiles, it is probably best to do this in the DB somehow (like the code snippet from above)

judtinzhang (commented):
Great first pass! Didn't look into the test cases, I'll trust you on that. A few comments on most files but most important comments in `user_stat.py`. Let's talk more about that file at the GBM.

backend/package-lock.json
rachllee: I'm actually also unsure- it appeared during my very convoluted env setup process but I will manually delete

backend/Pipfile
```diff
@@ -45,4 +45,4 @@ typing-extensions = "*"
 drf-excel = "*"
 
 [requires]
-python_version = "3"
```
judtinzhang: I'm not quite sure why the `Pipfile` and `Pipfile.lock` are changed. In general, if there are no dependency changes, we shouldn't touch these cause they're read directly from our production machines to install. Would suggest following [this](https://stackoverflow.com/questions/7147270/hard-reset-of-a-single-file) article to remove these changes from the PR

backend/ohq/statistics.py
```diff
@@ -232,3 +234,83 @@ def queue_calculate_questions_per_ta_heatmap(queue, weekday, hour):
         hour=hour,
         defaults={"value": statistic if statistic else 0},
     )
+
+
+def user_calculate_questions_asked(user):
+    num_questions = Question.objects.filter(asked_by=user).count()
+
+    if num_questions and not num_questions == 0:
```
judtinzhang: nit: I think you can just write `if num_questions:`. If `num_questions` is ever 0, the if statement would fail (0 is considered `False`)

backend/ohq/statistics.py
```diff
@@ -232,3 +234,83 @@ def queue_calculate_questions_per_ta_heatmap(queue, weekday, hour):
...
+def user_calculate_time_helped(user):
+    user_time_helped = (
+        Question.objects.filter(
+            asked_by=user,
+            status=Question.STATUS_ANSWERED
+            ).aggregate(time_helped=Sum(F("time_responded_to") - F("time_response_started")))
+    )
+    time = user_time_helped["time_helped"]
+
+    if time and not math.isclose(time.total_seconds(), 0, abs_tol=0.001):
```
judtinzhang, line 270: Question: is there a reason you use `time.total_seconds()` here, but a few lines down you use `time.seconds`? Is there a difference?

backend/ohq/management/commands/user_stat.py
```diff
@@ -0,0 +1,65 @@
...
+    user_calculate_time_helped,
+    user_calculate_time_helping,
+    user_calculate_students_helped,
+)
+from django.db.models import Q
+
+
+
+
+class Command(BaseCommand):
```
judtinzhang, line 22: Let's talk about this file on Friday!

judtinzhang (commented):
Almost there! Few nits

backend/ohq/statistics.py
rachllee: lol nope I don't think there's a difference I was just copying off of the documentation- I'll change it to .seconds for consistency

backend/Pipfile
rachllee: I tried to fix this with the stackoverflow article but it kind of does not look like it did anything per the last commit- I think I could use some help with this at GBM

judtinzhang: sounds good

backend/ohq/management/commands/course_stat.py
```diff
@@ -1,5 +1,6 @@
 from django.core.management.base import BaseCommand
 from django.utils import timezone
+import datetime
```
judtinzhang: nit: think we can remove this line

backend/ohq/management/commands/user_stat.py
```diff
@@ -0,0 +1,57 @@
...
+
+            for user in users_union:
+                    user_calculate_questions_asked(user)
+                    user_calculate_questions_answered(user)
+                    user_calculate_time_helped(user)
+                    user_calculate_time_helping(user)
+                    user_calculate_students_helped(user)
+
+
+    def handle(self, *args, **kwargs):
```
judtinzhang, line 44: looks a lot cleaner, good work!

judtinzhang (commented):
Almost there! Few nits. Please also run all the necessary linting for backend as well to format your code. I'm not sure how OHQ does it so please ask @trangiabach for that. One more detail as well that I'll message you about on slack

judtinzhang:
Nevermind, false alarm. Just that 1 nit and linting then we are good!

judtinzhang (approved):
LGTM!! Congrats on your first finished PR :D

danielxue (approved):
lgtm!

PR #313, opened 2024-04-19
Calendar frontend

danielxue:


judtinzhang:
hi dan

danielxue:
hi judtin

backend/Pipfile
```diff
@@ -10,15 +10,16 @@ flake8 = "*"
 flake8-absolute-import = "*"
 flake8-isort = "*"
 flake8-quotes = "*"
-django-debug-toolbar = "*"
+django-debug-toolbar = "3.2.4"
```
vcai122, line 13: what was the problem that made you have to fix the version

backend/ohq/serializers.py
```diff
@@ -525,27 +526,43 @@ def update(self, instance, validated_data):
             )
             or (
                 "rule" in validated_data
-                and (rule is None or rule.frequency != validated_data["rule"]["frequency"])
+                and (
+                    rule is None
+                    or validated_data["rule"] is None
+                    or rule.frequency != validated_data["rule"]["frequency"]
+                    or rule.params != validated_data["rule"]["params"]
+                )
```
vcai122, line 534: Very hard to understand what this is saying because it is not very pythonic. Seems like a comparison? Probably better and more maintainable to add a member method to the Rule class. 

Also in a lot of cases prefer `.get` as it allows a default value. e.g. `validated_data.get('rule', None)`. This way you don't need to explicitly check the `None` case each time.

backend/ohq/serializers.py
```diff
@@ -491,7 +491,7 @@ class EventSerializer(serializers.ModelSerializer):
     All times are converted to UTC+0
     """
 
-    rule = RuleSerializer(required=False)
+    rule = RuleSerializer(allow_null=True, required=False)
```
vcai122, line 494: is this allow_null for updates?

backend/ohq/serializers.py
```diff
@@ -525,27 +526,43 @@ def update(self, instance, validated_data):
...
+                    )
                 validated_data.pop("rule")
-            Occurrence.objects.filter(event=instance).delete()
 
+        if "rule" in validated_data:
+            validated_data.pop("rule")
         # can never change course_id, client should create a new event instead
         validated_data.pop("course_id")
+
         super().update(instance, validated_data)
```
vcai122, line 553: lol why is everything about modifying the instance being done manually in the serializer

backend/ohq/serializers.py
```diff
@@ -525,27 +526,43 @@ def update(self, instance, validated_data):
...
         ):
             if "rule" in validated_data:
-                rule, _ = Rule.objects.get_or_create(frequency=validated_data["rule"]["frequency"])
+                if validated_data["rule"] is None:
+                    rule = None
+                else:
+                    rule, _ = Rule.objects.get_or_create(
+                        frequency=validated_data["rule"]["frequency"],
+                        params=validated_data["rule"]["params"],
+                    )
```
vcai122, line 544: let's at least move the logic for actually updating the instance rule up here so its in the same place. 

This is setting the instance rule back to the initial one and saving if rule is not in validated_data. Let's be explicit about that, or better explicit and not do the extra db call with .save()

backend/ohq/serializers.py
```diff
@@ -525,27 +526,43 @@ def update(self, instance, validated_data):
...
+        Occurrence.objects.filter(event=instance).delete()
+
         return instance
 
     def create(self, validated_data):
         course = Course.objects.get(pk=validated_data["course_id"])
         rule = None
-        if "rule" in validated_data:
-            rule, _ = Rule.objects.get_or_create(frequency=validated_data["rule"]["frequency"])
+        if "rule" in validated_data and validated_data["rule"] is not None:
```
vcai122, line 564: same comment as above about .get() and walrus

backend/ohq/serializers.py
vcai122: also use walrus operator so you don't have to be redundant about accessing the dictionary.  
`rule_info := validated_data['rule']`

also unpack the dictionary: `frequency, params = validated_data['rule']['frequency'], validated_data['rule']['params']`

vcai122 (commented):
left some comment on the django/python. I don't actually know the business logic so can't really review the logic. Also is linting disabled?

danielxue:
Thanks for the review. There were some weird versioning issues with linting I think so we disabled it, there's a separate PR up to resolve that.

danielxue:
Merging before resolving some nits to launch before start of sem, will correct in future PR soon(TM)

PR #324, opened 2024-11-18
Add backend for bookable office hours

dkwncho:
- Created new Booking model, serializer, permission, urls, and viewset
- Bookings can only be created with start times of 5-minute intervals (e.g., :00, :05, :10, :15, etc), would need to reflect this on the frontend


backend/ohq/models.py
```diff
@@ -447,3 +451,25 @@ class Meta:
...
+    end = models.DateTimeField(_("end"), db_index=True, blank=True)
+
+    class Meta:
+        verbose_name = _("booking")
+        verbose_name_plural = _("bookings")
+        ordering = ["start"]
+        index_together = (("start", "end"),)
+
+    def __str__(self):
+        return gettext("%(start)s to %(end)s") % {
```
benjmnxu: it might be better to just use an f-string for this case. lmk if theres a specific reason u chose gettext.

backend/ohq/models.py
```diff
@@ -4,7 +4,11 @@
 from django.dispatch import receiver
 from email_tools.emails import send_email
 from phonenumber_field.modelfields import PhoneNumberField
-
+from schedule.models import Event, Occurrence
+from django.utils.translation import gettext, gettext_lazy as _
```
benjmnxu: not sure if we need to support translations.

backend/ohq/views.py
```diff
@@ -19,6 +19,8 @@
 from django.http import HttpResponseBadRequest, JsonResponse
 from django.utils import timezone
 from django.utils.crypto import get_random_string
+from django.utils.translation import gettext_lazy as _
```
benjmnxu: not sure we need these imports

backend/ohq/permissions.py
```diff
@@ -504,3 +504,69 @@ def has_permission(self, request, view):
...
+                    return True
+                else:
+                    updated_fields = request.data.keys()
+                    restricted_student_fields = ["occurrence", "start", "end"] # allowed_student_field would be "user"
+
+                    for field in updated_fields:
+                        if field in restricted_student_fields:
+                            return False
+                    if "user" in updated_fields:
+                        if booking.user is not None: # Doesn’t allow a student to update user if someone already booked the slot
```
benjmnxu: could just return booking.user is None

backend/ohq/views.py
```diff
@@ -772,6 +777,103 @@ def list(self, request, *args, **kwargs):
...
+    You must specify all of the fields or use a patch request.
+
+    partial_update:
+    Update certain fields in the Booking.
+    """
+
+    serializer_class = BookingSerializer
+    permission_classes = [BookingPermission | IsSuperuser]
+
+    def create(self, request, *args, **kwargs):
```
benjmnxu: are we only allowed to create bookings in chronological order? For example, if an occurrence is from 1-3pm and no other bookings existed, I would only be able to book a slot at 1pm. If someone had booked starting at 1pm, I would only be able to book a 1:30pm spot (assuming interval is 30mins).

benjmnxu (changes requested):
I think there is an issue with the Booking creation logic–please read my comment on it. Otherwise, I think we're pretty close to getting this merged. I'd encourage you to write some tests too.

backend/ohq/models.py
```diff
@@ -447,3 +449,179 @@ class Meta:
 
     def __str__(self):
         return f"{self.user}: {self.metric}"
+    
+class Occurrence(models.Model):
+    event = models.ForeignKey(Event, on_delete=models.CASCADE, verbose_name=("event"), related_name="occurrences")
+    title = models.CharField(("title"), max_length=255, blank=True)
```
vcai122: what is this ("title") thing, similar to on the other fields?

backend/ohq/models.py
```diff
@@ -447,3 +449,179 @@ class Meta:
...
+    start = models.DateTimeField(("start"), db_index=True)
+    end = models.DateTimeField(("end"), db_index=True)
+    cancelled = models.BooleanField(("cancelled"), default=False)
+    original_start = models.DateTimeField(("original start"))
+    original_end = models.DateTimeField(("original end"))
+    created_on = models.DateTimeField(("created on"), auto_now_add=True)
+    updated_on = models.DateTimeField(("updated on"), auto_now=True)
+    interval = models.IntegerField(("interval"), blank=True, validators=[MinValueValidator(5), MaxValueValidator(60)])
+
+    class Meta:
```
vcai122: does the default verbose names not work

backend/ohq/models.py
```diff
@@ -447,3 +449,180 @@ class Meta:
...
+    location = models.CharField(("location"), max_length=255, blank=True)
+    description = models.TextField(("description"), blank=True)
+    start = models.DateTimeField(("start"), db_index=True)
+    end = models.DateTimeField(("end"), db_index=True)
+    cancelled = models.BooleanField(("cancelled"), default=False)
+    original_start = models.DateTimeField(("original start"))
+    original_end = models.DateTimeField(("original end"))
+    created_on = models.DateTimeField(("created on"), auto_now_add=True)
+    updated_on = models.DateTimeField(("updated on"), auto_now=True)
+    interval = models.IntegerField(("interval"), blank=True, null=True)
```
vcai122: is interval supposed to represent the length of how long a student can book time for during this occurence? Should this be specified on the event then?

backend/ohq/models.py
```diff
@@ -447,3 +449,180 @@ class Meta:
...
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+
+        if self.interval is not None:
+            if self.pk: # If save is called on object update, not creation
+                self.bookings.all().delete()
+
+            delta = self.end - self.start
+            delta_minutes = delta.total_seconds() / 60
+            booking_count = int(delta_minutes // self.interval)
```
vcai122: I'm not a big fan of making booking objects automatically on save. I'm not totally against it since I see the value in just having the objects be there, but in some sense we are creating multiple sources of truths (obviously bad!!) --> i.e. we require the bookings to be synchronized with the interval field, hence why we are overriding save to make sure of this but this is generally kind of sus.

vcai122: For example maybe one way to avoid this is to initially have no bookings and allow users to create bookings, but you validate the time they specified (valid times can be derived from the start time and interval fields)

vcai122: Also not sure how events fit into the picture

vcai122: Lets talk next time I'm at gbm

backend/ohq/models.py
```diff
@@ -447,3 +449,180 @@ class Meta:
...
+    created_on = models.DateTimeField(("created on"), auto_now_add=True)
+    updated_on = models.DateTimeField(("updated on"), auto_now=True)
+    interval = models.IntegerField(("interval"), blank=True, null=True)
+
+    class Meta:
+        verbose_name = ("occurrence")
+        verbose_name_plural = ("occurrences")
+        index_together = (("start", "end"),)
+
+    def __init__(self, *args, **kwargs):
```
vcai122: i know you just copied it but why is this code like this 😭

backend/ohq/permissions.py
```diff
@@ -504,3 +508,50 @@ def has_permission(self, request, view):
...
+        if view.action in ["retrieve"]:
+            booking = Booking.objects.filter(pk=view.kwargs["pk"]).first()
+            membership = self.get_membership_from_event(request=request, event=booking.occurrence.event)
+            return membership is not None
+
+        if view.action in ["update", "partial_update"]:
+            booking = Booking.objects.filter(pk=view.kwargs["pk"]).first()
+            membership = self.get_membership_from_event(request=request, event=booking.occurrence.event)
+
+            if membership is not None:
```
vcai122: Can we use guard clauses instead of `else return False`

backend/ohq/permissions.py
```diff
@@ -504,3 +508,50 @@ def has_permission(self, request, view):
...
+                if membership.is_ta:
+                    return True
+                else:
```
vcai122: no need for else if you already returned True. A good heuristic for readable code is minimize indents

backend/ohq/permissions.py
```diff
@@ -504,3 +508,50 @@ def has_permission(self, request, view):
...
+                    updated_fields = request.data.keys()
+                    restricted_student_fields = ["occurrence", "start", "end"] # allowed_student_field would be "user"
+
```
vcai122: maybe something like `if any(restricted_field in updated_fields for restricted_field in restricted_fields): return False`.

More readable since you can just read it like english

backend/ohq/permissions.py
```diff
@@ -504,3 +508,50 @@ def has_permission(self, request, view):
             return True
 
         return True
+
+class BookingPermission(permissions.BasePermission):
```
vcai122, line 511: seems like you're always doing occurence.event so why not make this get_membership_from_occurence? also this is better as a static function

backend/ohq/migrations/0021_occurrence.py
joyliu-q, line 1: Can these be collapsed to one migration file?

backend/ohq/models.py
```diff
@@ -447,3 +448,70 @@ class Meta:
...
+        ordering = ["start"]
+        index_together = (("start", "end"),)
+
+    def clean(self):
+        if self.start >= self.end:
+            raise ValidationError('Start time must be before end time.')
+        
+        if self.start.minute % 15 != 0:
+            raise ValidationError('Start time must be on a 15-minute interval (e.g., :00, :15, :30, :45).')
+
```
joyliu-q: Can we document this behavior somewhere (either top level in code, in PR description, etc)

Also looping in @ryantanen who will likely be doing the frontend. If we are not allowing non-15-min intervals, we should also prevent booking non-15-min slots on the frontend as well.

joyliu-q (commented):
Would prefer if a backend looked at this but top level looks good to me.

backend/ohq/models.py
```diff
@@ -447,3 +448,70 @@ class Meta:
...
+        verbose_name = ("booking")
+        verbose_name_plural = ("bookings")
+        ordering = ["start"]
+        index_together = (("start", "end"),)
+
+    def clean(self):
+        if self.start >= self.end:
+            raise ValidationError('Start time must be before end time.')
+        
+        if self.start.minute % 15 != 0:
```
minghansun1: If start time needs to be in 15 minute intervals, should end time also be in 15 minute intervals? Otherwise there may be un-utilized time between the end of a booking and the start of the next one

minghansun1 (commented):
LGTM except the one comment i added

joyliu-q (approved):
Looks good to me!

PR #325, opened 2025-02-12
:bug: Bug Fix for Queue Settings Pin Toggle

ryantanen:
This is a minor bug fix to OHQ queue settings; there was an issue where you could not toggle off "Pin" and "Rate Limit" toggles once you turned them on. This was a small fix and works in testing.

Also, this patch was made using our new Waypoint tool, which is cool.

### DIFF
--- a/frontend/components/Course/InstructorQueuePage/QueueFormFields.tsx
+++ b/frontend/components/Course/InstructorQueuePage/QueueFormFields.tsx
@@ -172,11 +172,6 @@ const QueueFormFields = ({
         }
         setInput({ ...input });
     };
-
-    const handlePinInputChange = (e, { checked }) => {
-        input.pinEnabled = checked;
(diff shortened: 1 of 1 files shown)


frontend/components/Course/InstructorQueuePage/QueueFormFields.tsx
```diff
@@ -172,11 +172,6 @@ const QueueFormFields = ({
         }
         setInput({ ...input });
     };
-
-    const handlePinInputChange = (e, { checked }) => {
```
joyliu-q, line 176: Nit: Let's keep the `handlePinInputChange` function, but change the underlying implementation, so people reading this file have an easier time understanding.

joyliu-q (approved):
LGTM! Thanks Ryan

PR #327, opened 2025-02-16
uv migration

clay53:


clay53:
Rebased on #329 for testing purposes

clay53:
Rebased on #329 again. UV migrations for local development and the production-like docker image work now. Since django-base has not been updated, it is currently installing uv within the ohq container. uv and pipenv are still available for both dev and docker images. For docker images to use pipenv instead of the new default, uv, must set `DEPENDENCY_MANAGER` docker arg to `pipenv`. uv is the default so that will be used in prod without having to change and prod settings, but can be rolled back to use pipenv in case of failure.

backend/.dockerignore
```diff
@@ -15,3 +15,6 @@ LICENSE
 **/__pycache__/
 tests/
 postgres/
+
+# uv
```
joyliu-q: why is this uv specific? can we just include this in misc

backend/Dockerfile
```diff
@@ -1,12 +1,42 @@
 FROM pennlabs/django-base:b269ea1613686b1ac6370154debbb741b012de1a-3.11
+ARG DEPENDENCY_MANAGER="uv"
```
joyliu-q: can we default to not uv

backend/Dockerfile
```diff
@@ -1,12 +1,42 @@
...
 
 LABEL maintainer="Penn Labs"
 
+# Install build dependencies
+RUN apt-get update && apt-get install --no-install-recommends -y gcc libpq-dev libc-dev curl \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install uv
+# Download the latest installer
+ADD https://astral.sh/uv/0.6.0/install.sh /uv-installer.sh
```
joyliu-q: curious why is this ur installation method of choice?

backend/Dockerfile
```diff
@@ -1,12 +1,42 @@
...
+
+RUN echo "5464b06094b6363861b99cd60a010892e52fc2a7503b7594419828e74afecfe6  /uv-installer.sh" | sha256sum -c
+
+# Run the installer then remove it
+RUN sh /uv-installer.sh && rm /uv-installer.sh
```
joyliu-q: nit: any chance we can group the installation into a single run

backend/Dockerfile
```diff
@@ -1,12 +1,42 @@
...
+
+# Ensure the installed binary is on the `PATH`
```
joyliu-q: less comments, only at the top of sections when needed

backend/Dockerfile
```diff
@@ -1,12 +1,42 @@
...
+ENV PATH="/root/.local/bin/:$PATH"
+
 # Copy project dependencies
 COPY Pipfile* /app/
+COPY pyproject.toml /app/
+COPY uv.lock /app/
+
```
joyliu-q: can be collapsed

backend/Dockerfile
```diff
@@ -1,12 +1,42 @@
...
-RUN pipenv install --system
+# Install project dependencies if using uv if uv enabled
+RUN if [ "$DEPENDENCY_MANAGER" = "uv" ]; then uv sync --frozen --no-dev --no-install-project --python $(which python); fi
+# Make installed binaries available for POSIX compliant scripts
+ENV PATH="/app/.venv/bin:$PATH"
+# Patch for scripts that use a hard-coded path to uwsgi (django-run)
+RUN if [ "$DEPENDENCY_MANAGER" = "uv" ]; then  mv /app/.venv/bin/uwsgi /usr/local/bin/uwsgi; fi
+# Patch for scripts that use a hard-coded path to gunicorn (asgi-run)
+RUN if [ "$DEPENDENCY_MANAGER" = "uv" ]; then  mv /app/.venv/bin/gunicorn /usr/local/bin/gunicorn; fi
+# Patch for scripts that use a hard-coded path to python (django-run, asgi-run)
```
joyliu-q: i would prefer if we grouped it into uv and nonuv case for readability

joyliu-q (changes requested):
a few changes! thanks for the pr :D

backend/Dockerfile
clay53: I either misread the uv docs or it wasn't there (they seem to be pushing a lot of updates). I'm switching it to the simpler `COPY --from=ghcr.io/astral-sh/uv@sha256:2381d6aa60c326b71fd40023f921a0a3b8f91b14d5db6b90402e65a635053709 /uv /uvx /bin/`

joyliu-q (approved):
LGTM! Thanks

PR #328, opened 2025-02-28
Added health routes to backend and frontend

sritanmotati:
Created health page (health.tsx) in frontend
* If user-agent: service-status is a header, it will return OK
* Otherwise, it redirects to the home page
* `/health`

Created health route in backend
* `/api/health`


ryantanen (approved):
Looks good

PR #329, opened 2025-03-04
Add docker compose profile for emulating production

clay53:
Most important aspect of this is that it tests `django-run` and `asgi-run` are working. Not intended to replace development or production environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,3 +58,12 @@ Users from the populate script all have password `pennlabs`.
   - a. Navigate to `/frontend`
   - b. `yarn cypress open`
   - c. Click on `question.spec.ts`
+
+### Emulating the production server
+This will, as best as possible, emulate the production configuration described in `/k8s/main.ts`.
+
+1. Open a command prompt to `/backend`
+2. Run `docker compose --profile proddev up`
+3. To turn it off, use `CTRL+C` and then `docker compose --profile proddev down`
+
+Note this will reuse the postgres database stored in `/backend/postgres`. You may reset this database by turinging off the docker containers and deleting `/backend/postgres`
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ This will, as best as possible, emulate the production configuration described i
 
 1. Open a command prompt to `/backend`
 2. Run `docker compose --profile proddev up`
-3. To turn it off, use `CTRL+C` and then `docker compose --profile proddev down`
+3. Connect via `http://127.0.0.1:8000`
+4. To turn it off, use `CTRL+C` and then `docker compose --profile proddev down`
 
 Note this will reuse the postgres database stored in `/backend/postgres`. You may reset this database by turinging off the docker containers and deleting `/backend/postgres`
--- a/backend/docker-compose.yaml
+++ b/backend/docker-compose.yaml
@@ -15,6 +15,7 @@ services:
     image: redis:4.0
     ports:
       - "6379:6379"
+  # api, dashboard, and admin requests for proddev
   proddev-backend-wsgi:
     depends_on:
       - db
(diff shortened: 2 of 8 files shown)


backend/docker-compose.yaml
```diff
@@ -15,3 +15,53 @@ services:
...
+      - proddev
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - "8002:80"
+    environment:
+      - DATABASE_URL=postgres://postgres:postgres@db:5432/postgres
+      - DJANGO_SETTINGS_MODULE=officehoursqueue.settings.proddev
+    command: sh -c "python manage.py migrate && { DJANGO_SUPERUSER_PASSWORD=root python manage.py createsuperuser --no-input --username root --email root@pennlabs.org; /usr/local/bin/asgi-run; }"
```
joyliu-q, line 49: The createsuperuser command will fail if the database isn't ready yet. Instead, use a retry loop in the entrypoint

backend/docker-compose.yaml
```diff
@@ -15,3 +15,53 @@ services:
...
+  proddev-frontend:
+    profiles:
+      - proddev
+    build:
+      context: ../frontend
+      dockerfile: ../frontend/Dockerfile
+    ports:
```
joyliu-q, line 57: Nit: frontend local port should not be `8003`, even if its emulating production.

Using 8000 or 8080 is more standard because the frontend would not be an internal service.

Also I have a slight preference for `8080`, `8081`, and `8000` whenever possible and using other ports when these are exhausted.

backend/officehoursqueue/settings/proddev.py
```diff
@@ -0,0 +1,17 @@
+from officehoursqueue.settings.production import *
```
joyliu-q, line 3: Whats the reason for adding this new `proddev.py`

joyliu-q (changes requested):
Overall, a very good starting point! Added @ryantanen and @dkwncho, as well as @danielxue for posterity.

When adding new configs, it's good to specify a reason for that. Especially when changing the docker-compose.yaml, it's good to add in-line code comments on why this is needed. Same thing for top of `nginx-proddev.conf`.

Also not the biggest fan of the name `proddev`: if you can think of a better one would be cool

backend/docker-compose.yaml
clay53: The database will always be ready by the time that command runs right? It won't run if migrate fails

backend/docker-compose.yaml
clay53: Would
| service | port |
|- | - |
| nginx | 8000 |
| frontend | 8080 |
| backend-wsgi | 8081 |
| backend-asgi | 8082 |

work?

backend/officehoursqueue/settings/proddev.py
clay53: It's a basis for the django settings for proddev. It is based off of the production settings, but with some changes so it still actually works in the development environment. For example, unless `ALLOWED_HOSTS` is changed, it will return a 400 error on every request.

clay53: I don't believe it's possible to override Django settings with just environment variables without changing settings files.

backend/docker-compose.yaml
joyliu-q: yep!

joyliu-q (approved):
Theoretically lgtm. Would feel good about merging it as long as you tested it.

README.md
```diff
@@ -58,3 +58,13 @@ Users from the populate script all have password `pennlabs`.
   - a. Navigate to `/frontend`
   - b. `yarn cypress open`
   - c. Click on `question.spec.ts`
+
```
joyliu-q, line 61: Thanks for these docs!

backend/pyproject.toml
```diff
@@ -1,2 +1,60 @@
+[project]
+name = ""
```
joyliu-q, line 2: Could we add a project name here?

joyliu-q (commented):
Shaping up quite nicely! A small comment (feel free to ping me once u addressed) and should be good to go!

joyliu-q (approved):
lgtm

PR #330, opened 2025-03-16
Update shared actions to 0.1.7

clay53:
Bumps shared actions so they work after cache@v2 deprecation


clay53:
Note bump for publish-frontend's docker-publish isn't tested since change isn't not on `master`, but it worked for penn-courses: https://github.com/pennlabs/penn-courses/commit/68e03c6dd0be2167f2b458884d8e106599e348fc

clay53:
Need to bump to 0.1.7 instead

joyliu-q (approved):
LGTM

PR #332, opened 2025-03-30
Add backend for bookable office hours

dkwncho:
- Created new Booking model, serializer, permission, urls, and viewset
- Bookings can only be created with start times of 5-minute intervals (e.g., :00, :05, :10, :15, etc), would need to reflect this on the frontend


joyliu-q (approved):
LGTM! Please test everything before merging

